PR #3, opened 2017-02-23
Support GCC 6.3

Croydon:
This PR adds a Ubuntu 17.04 container with GCC 6.3.
Also it updates CMake to v3.7.2 in all containers.

In contrast to the other containers I run most of the commands in one RUN.
This reduces the Docker layers and therefore the size of the container.

Example build log can be found here https://hub.docker.com/r/croydon/conan-gcc6-3-docker-test/builds/b2ffwq6qne4baxvdbmcyh3d/

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,4 @@ The images are uploaded to Dockerhub:
 - [gcc 5.3](https://hub.docker.com/r/lasote/conangcc53/)
 - [gcc 5.4](https://hub.docker.com/r/lasote/conangcc54/)
 - [gcc 6.2](https://hub.docker.com/r/lasote/conangcc62/)
+- [gcc 6.3](https://hub.docker.com/r/lasote/conangcc63/)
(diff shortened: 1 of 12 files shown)


lasote:
Thanks so much. I'm regenerating the images now

Croydon:
@lasote Thank you! 

By the way, should I flood you now with pull requests for GCC6.3 builds of pkgs I would like to see or would this be more annoying than helping?

lasote:
For Boost I prefer to upload myself, I will make a commit for 6.3 only and then other with all builds that I will cancel. It takes like 24h to complete.
For other libraries, please, be my guest and open PRs. If I don't answer/accept the PRs in some time please write me in twitter, sometimes I just can't attend all the github notifications.

Croydon:
I created 9 pull requests.
Also you likely want to restart this job: https://travis-ci.org/lasote/conan-boost/jobs/204751192 :)

PR #8, opened 2017-06-02
Automatic docker push after build

uilianries:
Hi!

You will no longer need to execute the same push routine any more! I added some new scripts to build, test and push. So, test.sh, build.sh and run.sh are no longer needed.

Before to accept these changes, you need to configure DOCKER_PASSWORD in Travis, if you want upload some image.

You can see all results here: https://travis-ci.org/uilianries/conan-docker-tools/builds/238598032 and here are the images: https://hub.docker.com/r/uilianries/  
My last build has failed, because it is using lasote as docker username ;)

Saludos!


README.md
```diff
@@ -93,3 +93,45 @@ docker run -v/tmp/.conan:/home/conan/.conan lasote/conangcc63 bash -c "conan ins
...
+where you can choose compiler version, docker login and skip some stage.
+
+### Build
+The first stage collect all compiler versions listed and build their docker file.
+By default, the tag will be your ``docker_username/conan_compiler_version``.
+You can choose the compiler versions by environment variables or application arguments.
+If you do not set any compiler version, the script will execute all supported versions.
+
+### Test
+The second stage runs the new image and builds ``gtest/1.8.0`` with libstdc++, for x86 and x86_64. ``Clang`` images use libc++ and libstdc++.
```
lasote: Specifying the setting `compiler` and `compiler.version` in the build to ensure that the current installed compiler is correct.

conan_docker_tools/application.py
```diff
@@ -0,0 +1,71 @@
...
+        logging.info("Starting Conan Docker Tools")
+        self.__arguments = Arguments()
+        self.__variables = Variables()
+
+    def run(self):
+        gcc_versions, clang_versions = self.__get_compiler_versions()
+        self.__process_compiler(GccCompiler(), gcc_versions)
+        self.__process_compiler(ClangCompiler(), clang_versions)
+
+    def __process_compiler(self, compiler, versions):
```
lasote: For private methods we use only one "_" underscore by convention, it allows the IDE etc to access the code.

conan_docker_tools/compiler.py
```diff
@@ -0,0 +1,35 @@
+"""Static compiler information
+"""
+
+class Compiler(object):
```
lasote: NamedTuple("Compiler", "name, versions")

README.md
```diff
@@ -93,3 +93,45 @@ docker run -v/tmp/.conan:/home/conan/.conan lasote/conangcc63 bash -c "conan ins
...
+--no-build               Skip build stage
+--no-test                Skip test stage
+--conan-gcc-versions     GCC versions to build, test and deploy. This option is preferred over environment variable
+--conan-clang-versions   Clang versions to build, test and deploy. This option is preferred over environment variable
+```
+
+Supported variables:
+
+```
+CONAN_GCC_VERSIONS     GCC versions to build, test and deploy
```
lasote: In which format? [4.6, 4.9] ? [46,49] Provide an example

README.md
```diff
@@ -93,3 +93,45 @@ docker run -v/tmp/.conan:/home/conan/.conan lasote/conangcc63 bash -c "conan ins
 
 This command is sharing ``/tmp/.conan`` as a shared folder with the conan home, so the Boost package will be built there.
 You can change the directory or execute any other command that works for your needs.
+
+
+Build, Test and Deploy
+======================
+
+To create all default images, this project uses a bunch of script in conan_docker_tools. These scripts run as an application,
```
lasote: Introduce: The images are already built and uploaded to "lasote" dockerhub account, If you want to build your own images you can do it by...

conan_docker_tools/build.py
```diff
@@ -0,0 +1,14 @@
+"""Build Docker image
+"""
+import subprocess
+from image import Image
+
+
+class Build(object):
```
lasote: Replace this class with a simple function.

conan_docker_tools/compiler.py
```diff
@@ -0,0 +1,35 @@
...
+        """Retrieve supported versions
+        """
+        return self.__versions
+
+
+class GccCompiler(Compiler):
+    """Specilized builder for GCC
+    """
+    def __init__(self):
+        Compiler.__init__(self, "gcc", ["4.6", "4.8", "4.9", "5.2", "5.3", "5.4", "6.2", "6.3"])
```
lasote: The right way to call the parent constructor is:

super(GccCompiler, self).__init__(

conan_docker_tools/compiler.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class ClangCompiler(Compiler):
+    """Specialized builder for Clang
+    """
+    def __init__(self):
+        Compiler.__init__(self, "clang", ["3.8", "3.9", "4.0"])
```
lasote: super(ClangCompiler, self).init(

conan_docker_tools/deploy.py
```diff
@@ -0,0 +1,21 @@
+"""Deploy image to hub.docker
+"""
+from image import Image
+from variables import Variables
+import subprocess
+import os
+import sys
+
+
+class Deploy(object):
```
lasote: A class is not needed, just create a function

conan_docker_tools/test.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def _steps(self, image):
+        pass
+
+    @staticmethod
+    def make(image):
+        return TestGcc() if image.compiler == "gcc" else TestClang()
+
+
+class TestGcc(Test):
```
lasote: function test_gcc

conan_docker_tools/test.py
```diff
@@ -0,0 +1,49 @@
...
+    def _steps(self, image):
+        try:
+            subprocess.check_call("sudo docker exec %s conan install gtest/1.8.0@lasote/stable -s arch=x86_64 -s compiler=%s -s compiler.version=%s -s compiler.libcxx=libstdc++ --build" % (image.build_dir, image.compiler, image.version), shell=True)
+            subprocess.check_call("sudo docker exec %s conan install gtest/1.8.0@lasote/stable -s arch=x86 -s compiler=%s -s compiler.version=%s -s compiler.libcxx=libstdc++ --build" % (image.build_dir,  image.compiler, image.version), shell=True)
+        except:
+            raise
+
+
+class TestClang(TestGcc):
```
lasote: function test_clang

conan_docker_tools/image.py
```diff
@@ -0,0 +1,30 @@
+"""Docker image entity
+"""
+
+from variables import Variables
+
+
+class Image(object):
```
lasote: Can be replaced with a NamedTuple and just a "get_variables" function

lasote:
Hi!! thanks for your help.

In general, the PR is ok, my main concerns are:

- In python we always try to avoid boilerplate classes, in Java, your PR would be perfect, but in python is very very verbose. We try to avoid the "bean" clases with private attributes and properties, if somone is going to access the property just keep them public at least there is a critical reason to not allow the write.
- Most of the classes can be replaced with tiny functions. 
- In the .travis.yml we need to introduce some check like TRAVIS_BRANCH == "master" to avoid upload images with a PR (anyone could override my images if I set the password). If don't just don't run the upload, but not fail, we need to check if the PR is ok.
- All the arguments you can pass to the program to generate the classes, maybe is oversized, as a principle we like to go step by step and only develop a feature when it's really needed. Do you think that many people are going to build the images by their own and use these parameters? I don't.
- For that reason, I think we should simplify the code, maybe keep the latest version and just introduce the possibility of upload the images (just with a new function).

Thanks for all your efforts! 👍

uilianries:
Thanks for the review and comments!

I'll rectify ASAP.

Regards.

uilianries:
Done! @lasote 

I simplified all to build.py.

I just removed DOCKER_USERNAME from Travis file to enable anyone to create PR without errors.

The conan-package-tools was a good project to take a look. I brought CONAN_STABLE_BRANCH_PATTERN from there.

Now, you will need to include DOCKER_USERNAME and DOCKER_PASSWORD in your Travis settings, but this is documented.

Have a nice weekend!

build.py
```diff
@@ -0,0 +1,136 @@
...
+            return
+
+        current_branch = subprocess.check_output(['git', 'name-rev', '--name-only', 'HEAD']).decode()
+        prog = re.compile(self.variables.conan_stable_branch_pattern)
+        if not prog.match(current_branch):
+            logging.warning("Skipped upload, stable branch pattern does not match!")
+            return
+
+        logging.info("Login to Docker hub account")
+        result = subprocess.call(['sudo', 'docker', 'login', '-p', self.variables.docker_password, '-u', self.variables.docker_username])
```
lasote: Maybe avoid sudo and append the sudo in the .travis.yml?

build.py
```diff
@@ -0,0 +1,136 @@
...
+        if result != os.EX_OK:
+            raise RuntimeError("Could not login username %s to Docker hub." % self.variables.docker_username)
+
+        logging.info("Upload Docker image %s to Docker hub." % image_name)
+        subprocess.check_call("sudo docker push %s" % image_name, shell=True)
```
lasote: Idem

lasote:
Now It looks very good! much more pythonic ;) Good job.
About the "sudo" in commands, maybe we should remove it and append it to the .travis / python run. For example in OSX is not needed to use sudo in docker commands.

uilianries:
Good point, I'm remove.

uilianries:
@lasote Done!

lasote:
Hi, I'm taking a look to this and finally see if we can improve the automatic creation of the images.

lasote:
I've simplified a couple things (no more default lasote and renamed image env vars, no more default images), added the conan_server image building:

https://github.com/conan-io/conan-docker-tools/pull/19

Let's see how it goes. All your commits are there, so you are the main contributor of this feature ;)

PR #10, opened 2017-08-06
Support gcc-7

uilianries:
Hi!

I bring a new Docker image based on Ubuntu Artful and gcc-7.  Ubuntu Artful is under devel, but gcc-7 is available by default on apt-get.    
Also, I updated CMake version to 3.9.0 for all images.

Now we can provide support for C++17 on Travis


lasote:
👍 👍 👍 👍 👍

lasote:
I'm going to generate the images and upload them. Thanks so much!

PR #15, opened 2017-10-25
Update conan_server version in Dockerfile

ovidiub13:


lasote:
The problem here is that we've not merged yet this PR and it's already outdated...
I agree we should improve this, but it's almost mandatory to upgrade conan when you run an image.

lasote:
I've created a PR with this and some other #17

ovidiub13:
I think you gave the wrong PR number. I could have updated the version in this PR and get it through.

PR #17, opened 2017-10-25
Fix gcc 7.x Docker image test

ovidiub13:
Gcc-6 is not installed. This throws a `file not found` error, so I'm proposing to remove it.


ovidiub13:
In Ubuntu Artful there is no official package for GCC 7.1 anymore, just GCC 7.2.
I've updated all references to 7.2.

lasote:
But then we are losing the 7.1 image... Breaking existing builds that used it. We should somehow force install 7.1 in an image

ovidiub13:
@lasote I've tried. The package does not exist in the Ubuntu repository.

lasote:
Yeah, I saw it too. It makes no sense, thanks for open this discussion. It's on my last PR included.

PR #19, opened 2017-10-30
Feature/ci complete

lasote:


lasote:
It fails because of #17. As I commented we can do two things:

- Forget about the old 7.1 and only support 7.2 (and the latest if it keeps happening the same), so the already uploaded images (7.1 in this case) will be there but only "frozen" and with no maintenance anymore, only binary support, no more Dockerfile, no more possibility of an upgrade.
- Try to force somehow the distribution to install 7.1 or downgrade, I've done it in the past with some other images, but it requires time to do it.

@uilianries @ovidiub13 WDYT? I think we should provide the 7.1 somehow.

uilianries:
I prefer the first one.

We have a bunch of images to maintain and in many it's a headache try fix distro environment. But my opinion is, we will spend to much effort to fix just another version, meanwhile we should care about 7.2 because it's the latest.

Also, we could provide a table in README file with all images status (build and life cycle).

ovidiub13:
@lasote What's the purpose of these images?

* To provide what the disto offers + conan, or do we need a specific compiler? If the latter there are [official `GCC` Docker images](https://hub.docker.com/_/gcc/) based on Debian. They also skip 7.1.
* To provide specific tools that we consider common, and make it easier for users?

Getting the Ubuntu folks to also ship 7.1 will be rather difficult.

lasote:
Ok. I've talked with @memsharded and it makes sense since >5 where gcc minor is like a patch of <5. Probably we will take into consideration for conan/settings too. We are thinking about it.

lasote:
Related with: @ovidiub13 https://github.com/conan-io/conan/issues/1214
So we are deprecating the 7.1 image as you proposed. Probably we will do something with the conan settings, probably it makes no much sense to keep using the minor version by default for gcc 7.
@memsharded is opening an issue in conan repository.

README.md
```diff
@@ -7,20 +7,28 @@ You can use these images directly in your project or with the [conan-package-too
...
+| [lasote/conangcc46: gcc 4.6](https://hub.docker.com/r/lasote/conangcc46/)   |  Supported |
+| [lasote/conangcc48: gcc 4.8](https://hub.docker.com/r/lasote/conangcc48/)   |  Supported |
+| [lasote/conangcc49: gcc 4.9](https://hub.docker.com/r/lasote/conangcc49/)   |  Supported |
+| [lasote/conangcc52: gcc 5.2](https://hub.docker.com/r/lasote/conangcc52/)   |  Supported |
+| [lasote/conangcc53: gcc 5.3](https://hub.docker.com/r/lasote/conangcc53/)   |  Supported |
+| [lasote/conangcc54: gcc 5.4](https://hub.docker.com/r/lasote/conangcc54/)   |  Supported |
+| [lasote/conangcc62: gcc 6.2](https://hub.docker.com/r/lasote/conangcc62/)   |  Supported |
+| [lasote/conangcc63: gcc 6.3](https://hub.docker.com/r/lasote/conangcc63/)   |  Supported |
+| [lasote/conangcc71: gcc 7.1](https://hub.docker.com/r/lasote/conangcc71/)   |  DEPRECATED, only frozen binary image support, No Dockerfile |
+| [lasote/conangcc72: gcc 7.2](https://hub.docker.com/r/lasote/conangcc72/)   |  Supported |
```
uilianries: Very nice!

PR #22, opened 2017-11-10
Hotfix for GCC 7.2

uilianries:
Hi!

* After to install g++-7.2 the system doesn't solve c++ and g++ default names, so I used *update-alternatives* for that. 
* Also, when we need to build a C++ package for x86, the header *asm/error.h* is not found. I added *linux-libc-dev:i386* to solve that
* To make sure for next changes, I added a c++ package build (GTest). Unfortunately, conan-center doesn't provide any C++ package, only header only or Poco, but Poco is huge for a simple test. So, I added conan-community. 

Regards

### DIFF
--- a/build.py
+++ b/build.py
@@ -105,16 +105,6 @@ class ConanDockerTools(object):
                                   "-s compiler.libcxx=libstdc++ --build" %
                                   (container_name, compiler_name, compiler_version), shell=True)
 
-            if compiler_name == "clang":
-                subprocess.check_call("docker exec %s conan install zlib/1.2.11@conan/stable "
-                                      "-s arch=x86_64 -s compiler=%s -s compiler.version=%s "
-                                      "-s compiler.libcxx=libstdc++ --build" %
-                                      (container_name, compiler_name, compiler_version), shell=True)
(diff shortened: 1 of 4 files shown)


build.py
```diff
@@ -91,6 +91,20 @@ def test(self, compiler_name, compiler_version, image_name):
...
+            subprocess.check_call("docker exec %s conan install gtest/1.8.0@conan/stable -s "
+                                  "arch=x86_64 -s compiler=%s -s compiler.version=%s "
+                                  "-s compiler.libcxx=libstdc++ --build" %
+                                  (container_name, compiler_name, compiler_version), shell=True)
+            subprocess.check_call("docker exec %s conan install gtest/1.8.0@conan/stable "
+                                  "-s arch=x86 -s compiler=%s -s compiler.version=%s "
+                                  "-s compiler.libcxx=libstdc++ --build" %
+                                  (container_name, compiler_name, compiler_version), shell=True)
+
             if compiler_name == "clang":
```
uilianries: @lasote I didn't understand why clang builds are replicated at this point.

lasote: ? Neither do I, maybe we can remove it?

uilianries: Done!

PR #24, opened 2017-11-14
Build and Test using Docker Compose

uilianries:
Hi!

This PR brings the support to use Docker compose!

Also, It's possible to configure the tag name by `DOCKER_BUILD_TAG`  
By default, the tag **latest** will be used, but you can configure as env var, e.g. 0.28.1  
There is no way to apply multiple tags using docker-compose for now.

So, anyone will be able to build any image by builld.py or directly invoking docker-compose.

Regards!


uilianries:
I need to change a new point before to merge!

Issue #32, opened 2018-01-30
Update docker images


uilianries:
Hi!

@solvingj reported about an error during [Boost Python build](https://travis-ci.org/bincrafters/conan-boost_python/jobs/335001227#L1954). As you can see, we have distribution support error.

After that, I read [Ubuntu EOF page](https://www.ubuntu.com/info/release-end-of-life). Well, zesty (17.04) is dead.

We need to update all images to prevent EOF. IMO, would be nice to use Ubuntu 18.04 as base.

Regards!

solvingj:
This may also be affecting SDL2 and zmq packages now.

solvingj:
This is good, because it is already starting to hold up boost because it's causing `boost_python` to fail,

uilianries:
Fixed by #33

Issue #31, opened 2018-01-29
Docker Images for x86


uilianries:
Hi!

As we know, all images support both x86 and x86_64 builds. However, when we need to install some system package that is not a Conan package yet or is very distro dependent, we need to run SystemPackageTools.

Unfortunately, many packages are not prepared to be installed on same system using arch variation. For example, when I need to install some amd64 deb package on Ubuntu, the same header directories are used by i386 version. As result, we have a conflict during install. 

As workaround, we append the target on package name, please check our [SDL2](https://github.com/bincrafters/conan-sdl2/blob/stable/2.0.7/conanfile.py#L61) recipe to understand better. This works well when we have only one arch build by job.

Would be nice to provide recipes without this workaround, including their package. But, to have such situation, we need to create new Docker images, based on 32-bits arch. Fortunately, we have official docker images for Ubuntu 32-bits on [hub.docker](https://hub.docker.com/r/i386/ubuntu/)

My suggestion to solve such problem, is implement 32 bits version of Conan docker images. I know that this will increase (duplicate) the number of images, but will provide a way to build our recipes with no workarounds to solve multi-arch limitations. As done before, I could help to implement such images.

Regards!

/cc @solvingj

lasote:
Sure, support images for 32 bits and in general any combination that can help to develop, generate or test packages is always welcome. Your help as always, very appreciated.

uilianries:
Thanks! I'll start this task soon.

solvingj:
awesome work!!

PR #33, opened 2018-01-30
Fix Clang images (#32)

uilianries:
Hi!

I updated all clang images to use Ubuntu 17.10 (artful) as base to fix Zesty EOL. Also, artful will end in the middle of this year.

I tried to use Bionic, but there is an error when it try to install lib32stdc++6. Bionic is under development, so I want to update all images in April.

Also, I just removed all package versions because we need to fix clang and put Bincrafters back on the rails. But I promise that I will restore package versions during #31 

Regards!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ The images are uploaded to Dockerhub:
 | [lasote/conangcc54: gcc 5.4](https://hub.docker.com/r/lasote/conangcc54/)   |  Supported |
 | [lasote/conangcc62: gcc 6.2](https://hub.docker.com/r/lasote/conangcc62/)   |  DEPRECATED, only frozen binary image support, No Dockerfile |
 | [lasote/conangcc63: gcc 6.3](https://hub.docker.com/r/lasote/conangcc63/)   |  Supported |
+| [lasote/conangcc64: gcc 6.4](https://hub.docker.com/r/lasote/conangcc64/)   |  Supported |
 | [lasote/conangcc71: gcc 7.1](https://hub.docker.com/r/lasote/conangcc71/)   |  DEPRECATED, only frozen binary image support, No Dockerfile |
 | [lasote/conangcc72: gcc 7.2](https://hub.docker.com/r/lasote/conangcc72/)   |  Supported |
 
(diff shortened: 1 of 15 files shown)


uilianries:
Ops!

We have some problem to build GTest: 

/usr/bin/../lib/gcc/x86_64-linux-gnu/7.2.0/../../../../include/c++/7.2.0/bits/std_abs.h:101:3: error: unknown type name '__float128'

GTest is not official for now, so I'll update to use another c++ package from Conan center.

uilianries:
Forget about! I just retrieved Ubuntu vault to sources.list for now.

I want to wait for 18.04 and update all images as necessary.

uilianries:
I removed conangcc62 in favor of conangcc6.

The real motivation is that image is based on Ubuntu yakkety (EOL). Unfortunately (or not), new Ubuntu versions doesn't  support gcc-6.2, only >6.3

Also, I added information about conangcc5, conangcc6 and conangcc7 on README.

uilianries:
There an error running x86 builds. I'm adding a new test to check this problem.

README.md
```diff
@@ -16,19 +16,28 @@ The images are uploaded to Dockerhub:
 | [lasote/conangcc52: gcc 5.2](https://hub.docker.com/r/lasote/conangcc52/)   |  Supported |
 | [lasote/conangcc53: gcc 5.3](https://hub.docker.com/r/lasote/conangcc53/)   |  Supported |
 | [lasote/conangcc54: gcc 5.4](https://hub.docker.com/r/lasote/conangcc54/)   |  Supported |
-| [lasote/conangcc62: gcc 6.2](https://hub.docker.com/r/lasote/conangcc62/)   |  Supported |
+| [lasote/conangcc62: gcc 6.2](https://hub.docker.com/r/lasote/conangcc62/)   |  DEPRECATED, only frozen binary image support, No Dockerfile |
 | [lasote/conangcc63: gcc 6.3](https://hub.docker.com/r/lasote/conangcc63/)   |  Supported |
```
lasote: 6.4 missing

docker-compose.yml
```diff
@@ -70,6 +70,13 @@ services:
         image: "${DOCKER_USERNAME}/conangcc63:${DOCKER_BUILD_TAG}"
         container_name: conangcc63
         tty: true
+    conangcc64:
```
lasote: So conangcc62 is not deprecated?

uilianries: I just removed from travis file, but you are right, it doesn't make any sense keep on docker-compose. I'll remove it.

uilianries:
Done!

PR #34, opened 2018-02-04
Docker images for x86 (#31)

### LINKED ISSUE #31: Docker Images for x86

uilianries:
Hola!

This PR includes:

- Supported docker images for x86
  - gcc 4.9, 5, 6, 7
  - clang 3.9, 4.0, 5.0
- Updated zesty images to artful. It solves EOL package support
- Updated arch information on README

All x86 images has the suffix **-i386** and all x86_64 (multi-arch) are not changed.

Saludos!

Closes #31


lasote:
Amazing, many thanks

Issue #43, opened 2018-05-02
Support GCC 8


uilianries:
Hi!

Would be nice to provide support for new GCC version:

https://gcc.gnu.org/gcc-8/changes.html

I didn't check yet, but maybe Ubuntu 18.04 supports it. If not, we could build this version on latest Ubuntu image.

Regards.

uilianries:
Done!

PR #45, opened 2018-05-22
Add GCC 8 support

uilianries:
Hi!

This PR brings the support to use GCC8 + Conan.

All gcc8 jobs failed on Travis, because Conan 1.3.3 has no support for GCC8 in settings.yml

As a plus, I updated all CMake to latest stable version (3.11.2)

Related issue: #43


lasote:
This is great! I would prefer to release this after 1.4 (by the end of the month) to avoid issues.
Many thanks!

uilianries:
I agree. Otherwise, we need to provide some workaround to push gcc 8 in settings.yml.

uilianries:
Sure! I'm restating now!

uilianries:
Well, there is a bug on GCC-x86. Give some minutes to submit a patch. The image doesn't find the cmake command.

uilianries:
There are some considerations:

Both images lasote/conangcc8armv7 and lasote/conangcc8armv7hf depends on lasote/conangcc8 to build. So, I need to update the job to build all dependencies before. I'll solve this later.

uilianries:
@lasote Done!

Also, I fixed GCC 4.9

gcc_4.9-x86/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM i386/ubuntu:trusty
+FROM i386/ubuntu:xenial
```
lasote: I need to keep 4.9 on trusty because of the issues with the glibc. I changed it some commits ago, I don't know if you remember:

https://github.com/conan-io/conan-docker-tools/commit/60c4f542929580f4bcdbb44b6564db105351c762

gcc_4.9/Dockerfile
```diff
@@ -1,54 +1,56 @@
-FROM ubuntu:trusty
+FROM ubuntu:xenial
```
lasote: Same here

gcc_4.9-x86/Dockerfile
uilianries: understood.

uilianries:
Done! I just reverted GCC 4.9 but also updated CMake package.

PR #49, opened 2018-07-25
Update CMake & EOL Ubuntu versions

Croydon:
This
  * updates end-of-life Ubuntu versions to the next reasonable supported Ubuntu version
    * Artful -> Bionic 
    * Precise -> Trusty
    * This wasn't possible for all EOF Ubuntu based containers; some minor compiler versions aren't available in the newer package repositories; these containers are unchanged
  * updates CMake to 3.12.0


lasote:
Hi, The distros are chosen in purpose, when you update a distro you update the `glibc` and the binaries built against the new `glibc` can be (and will be) incompatible with the previously generated binaries.
We try that every gcc version comes with the "native" distribution that supported it.
About cmake, probably you know it but you can use the `https://github.com/conan-io/conan-docker-tools/blob/master/update_cmake.py` to replace it.

Issue #53, opened 2018-09-10
Use python 3.6 for conan docker images


db4:
Conan docker images use python 2.7 as a default interpreter. Python 2 is going to be [deprecated soon](https://docs.conan.io/en/latest/installation.html#python2) by conan, maybe it's time to switch to 3.6?

uilianries:
I agree!

uilianries:
I'll do it :muscle:

lasote:
Released

PR #63, opened 2018-10-17
Run Python 3 by default (#53)

uilianries:
- Use only official packages (conan center) during the tests
- Add validation for python and pip version
- Add custom logo
- Update conan server image (conan 1.8.3)
- Install python 3 in all images
- Update alternatives to run python as python3
- Update gcc 4.9-arm images avoiding x86 packages

UPDATE:

- Install python by pyenv
- Run python3 by default

rip python 2


Croydon:
My concern with this is that it uses some Python 3 minor versions which aren't supported anymore. Is there a way to get 3.6 or 3.7 for all?

uilianries:
it's the same problem related to other packages. we could add python 3.7, but there are some limitations.

Using official apt source list (apt install python-3.7-dev) they don't work properly, some sigfault is raised during the tests.

We could add extra source list (python-software-properties), but it requires python 2 and a bunch of extra packages.

Or we could download python3.7 source code, build and install.

Conan doesn't require python 3.7 so we could used python 3.4 I think.

lasote:
Yeah, would be nice to have > 3.6 but I understand the limitations. Conan test-suite is running with python 3.4 (only when we create a release/master) so it should work well.
@conan-io/barbarians team are we ready for this?

tru:
Have you guys considered pyenv? It's what we use in our docker images to get the latest versions of python.

Croydon:
pyenv sounds great. Bincrafters are already using it on macOS

https://github.com/bincrafters/conan-templates/blob/f241593bd9aed80b3bb729baea031c899b2f489d/.ci/install.sh#L16

Croydon:
IIRC correctly Python 3.7 works only on Ubuntu Xenial and newer, so I would suggest to install 3.6 on all older containers and 3.7 starting with Xenial. (e.g. https://github.com/travis-ci/travis-ci/issues/9069)

tru:
pyenv would allow each project to select python version as well - by adding a .python-version file in the root of the project.

uilianries:
Just sharing:

```docker
RUN mkdir -p /home/conan/.conan \
    && wget --no-check-certificate --quiet -O /tmp/pyenv-installer https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer \
    && chmod +x /tmp/pyenv-installer \
    && /tmp/pyenv-installer \
    && sudo ln -s /home/conan/.pyenv/bin/pyenv /usr/local/bin/pyenv \
    && echo 'eval "$(pyenv init -)"' >> ~/.bashrc \
    && echo 'eval "$(pyenv virtualenv-init -)"' >> ~/.bashrc \
    && rm /tmp/pyenv-installer \
    && pyenv install 3.7.0 \
    && pyenv global 3.7.0
```

The context above install pyenv and set python 3.7 as default. However only conan user is able to see python command, sudo will not find python command.

So we could update CPT to skip sudo command when updating conan and cpt on docker runner.

uilianries:
Thanks for sharing. IMO we could assume conan user (non-root) by default. Run any command as root in our images sounds like edge case.

Croydon:
is pyenv installing pip and virualenv as well? I see that you removed the current way we installed it

uilianries:
Yes. Pyenv is amazing!

uilianries:
Done!

All images are running python3 by pyenv.

Only docker images for GCC 4.x are running python 3.6, all others are running python 3.7

README.md
```diff
@@ -131,7 +133,7 @@ Use the images locally
 You can also use the images locally to build or test packages, this is an example command:
 
 ```
-docker run -v/tmp/.conan:/home/conan/.conan conanio/gcc63 bash -c "conan install zlib/1.2.11@conan/stable --build missing"
+docker run -v/tmp/.conan:/home/conan/.conan conanio/gcc63 bash -c "conan install lz4/1.8.3@bincrafters/stable --build missing"
```
Croydon: Why this change?

uilianries: Good question. I started this branch days ago when we had a problem downloading zlib from bintray. The goal is the same, build a C project. LZ4 is smaller than zlib thus our tests will be faster than before

Croydon:
Azure failed

> azure-pipelines.yml Expected document start parse event

lasote:
Merge planned for next week, on Monday probably. Ping me during the day.

PR #72, opened 2018-11-02
#12 Tag Docker images by Conan version

### LINKED ISSUE #12: Create tag by Conan version
Hi!

Conan 0.26 is out, but all images were created in version 0.25 (I guess). We can update conan by pip, but would be nice if the image brings the latest version ever, as lasote/conangcc63:latest. Also, would be nice to use a tag per version:

```
REPOSITORY                      TAG
lasote/conangcc63               latest
lasote/conangcc63               0.26.0
lasote/conangcc63               0.25.0
...
```
Regards.

uilianries:
- Add docker tag command to create latest and image version
- Move env vars to Travis settings
- Build Conan server

closes #12


Croydon:
Nice, after that we only need a trigger 
new conan version -> new cdt build
and we have always up to date containers and don't need to install/update Conan in every single job

uilianries:
it would be nice but I think pip packages are released manually so new tag version won't be enough

build.py
```diff
@@ -158,6 +182,19 @@ def deploy(self, service):
 
         logging.info("Upload Docker image from service %s to Docker hub." % service)
         subprocess.check_call("docker-compose push %s" % service, shell=True)
+        image_name = "%s/%s:%s" % (self.variables.docker_username, service, client_version)
+        logging.info("Upload Docker image %s" % image_name)
+        subprocess.check_call("docker push %s" % image_name, shell=True)
+
+    def tag(self, service):
```
lasote: wrong indent

uilianries: Applied PEP8

uilianries:
> And what if we push the images without changing the conan version? Can the tags be overwritten? Should we?

Yes, tags can be overwritten. The only problem is latest tag, where will be pointing to conan version.

Maybe we could run Travis monthly to update docker images (cron feature). As Conan releases one version by month we could start the job on the last day.

PR #83, opened 2018-11-30
CentOS image for ancient glibc (#82)

uilianries:
CentOS6 + gcc7 + glibc 2.12 + Conan?

Yes, we have!

Please people, review this PR.

/cc @Croydon 

closes #82

PR #84, opened 2018-12-06
Allow centos image install pip packages without sudo

uilianries:
#### Description

To keep the same behavior between Ubuntu images and CentOS.

The user **conan** was added to **wheel** group and now all python base path owned by **wheel** group, which means both **root** and **conan** are able to install python packages.

#### Issues
- #83


uilianries:
Conan server job is broken, but is related to https://github.com/conan-io/conan/issues/4080

PR #88, opened 2018-12-14
#87 Install conan package tools

### LINKED ISSUE #87: build with CONAN_DOCKER_IMAGE_SKIP_UPDATE=1 fails
I tried to build ninja_installer with `CONAN_DOCKER_IMAGE_SKIP_UPDATE=1` as documentation suggests.
currently, there is no point wasting time on conan update, as CDT rebuilt after conan or CPT releases.
unfortunately, it fails with the following error message:
```
/bin/sh: run_create_in_docker: command not found

```
full log at https://travis-ci.com/bincrafters/conan-ninja_installer/jobs/165005420

/cc @uilianries

uilianries:
When CONAN_DOCKER_IMAGE_SKIP_UPDATE is declared, all pip steps are skipped, however, this includes cpt installation.

closes #87

### DIFF
--- a/clang_3.8/Dockerfile
+++ b/clang_3.8/Dockerfile
@@ -78,7 +78,7 @@ RUN dpkg --add-architecture i386 \
     && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.7.1 \
     && pyenv global 3.7.1 \
     && pip install -q --upgrade --no-cache-dir pip \
-    && pip install -q --no-cache-dir conan \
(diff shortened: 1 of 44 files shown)


lasote:
This can be a problem. When we release Conan, CDT build is launched, but CPT is usually not released yet. So it could happen that there is no CPT compatible with Conan and the build will fail.
Probably we would need to release CPT first. I'm talking with the team.

uilianries:
We could build CDT for every CPT release. Maybe using Github hooks to avoid manual triggering.

Issue #89, opened 2018-12-15
reduce size of docker images - review installed software


SSE4:
currently provided docker images are pretty heavyweight. for instance, `conanio/gcc8` currently takes 1,03Gb. as docker images are pulled on each build, it introduces additional delay on build job, which scales linearly based on number of travis jobs. I think it makes a priority to keep our docker images as small as possible. I have reviewed `conanio/gcc8` image a bit (via [ncdu](https://dev.yorhel.nl/ncdu) program - `sudo ncdu /` command) and spot the following:

- `valgrind` - `/usr/lib/valgrind` takes **63,7 Mb**. I believe valgrind is usually used only for memory profiling, is it really necessary to include it into build images?
- `vim` -  `/usr/share/vim` takes **31 Mb**. also, do we need text editor, especially such a heavy-weight as part of build images?
- `cmake` - obviously, CMake itself is required, no doubts here. however, `/usr/bin/cmake-gui` is **24,7 Mb,** plus `/doc/cmake` is **24,2 Mb**, plus `/usr/share/cmake-3.12/Help` is **6,3 Mb**. I don't think we need GUI and documentation for CI builds, right?
- `python` - first off, we install python 3.7, but there is python 3.6 installation (seems like it's dependency of some other package) - not sure if we can avoid having two pythons? second, `/opt/pyenv/versions/3.7.1/lib/python3.7/test/__pycache__` takes **32Mb** - I doubt we need it at all. there are few other `__pycache__` directories around.

from approximate calculation, removing the above may easily save **~200Mb** per image (which is 20% of size).

NOTE: docker pull time on travis is currently around 30-40 seconds.

/cc @lasote @uilianries @danimtb @jgsogo @memsharded

uilianries:
Amazing investigation! I totally agree with you!

> valgrind - /usr/lib/valgrind takes 63,7 Mb. I believe valgrind is usually used only for memory profiling, is it really necessary to include it into build images?

Maybe we could provide valgrind_installer package instead.

> vim - /usr/share/vim takes 31 Mb. also, do we need text editor, especially such a heavy-weight as part of build images?

No. Our images were created to build packages only. Also all images have `vi` installed.

[...]

PR #92, opened 2018-12-22
Reduce image size phase1

SSE4:
fixes https://github.com/conan-io/conan-docker-tools/issues/89

- output image size during the build
- do not install valgrind
- do not install vim (this also avoids python 3.6 installation)
- exclude CMake GUI & docs during untargz
- remove `__pycache__` directories

for GCC 6 image, effect is: `1.02 GB` -> `734.36 MB`

/cc @uilianries @lasote @memsharded @jgsogo @danimtb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ env:
     - BUILD_CONAN_SERVER_IMAGE=1
 
 install:
-  - pip install conan
+  - pip install conan humanfriendly
 
 script:
   - python build.py
--- a/clang_3.8/Dockerfile
+++ b/clang_3.8/Dockerfile
@@ -57,6 +57,9 @@ RUN dpkg --add-architecture i386 \
     && rm -rf /var/lib/apt/lists/* \
     && wget -q --no-check-certificate https://cmake.org/files/v3.12/cmake-3.12.1-Linux-x86_64.tar.gz \
     && tar -xzf cmake-3.12.1-Linux-x86_64.tar.gz \
+       --exclude=bin/cmake-gui \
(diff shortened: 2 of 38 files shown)


clang_3.8/Dockerfile
```diff
@@ -80,6 +81,7 @@ RUN dpkg --add-architecture i386 \
     && pip install -q --upgrade --no-cache-dir pip \
     && pip install -q --no-cache-dir conan conan-package-tools \
     && chown -R conan:1001 /opt/pyenv \
+    && find /opt/pyenv -iname __pycache__ -print0 | xargs -0 rm -rf \
```
danimtb: Find this a little bit tricky. Maybe it will benefit from a comment

.travis.yml
```diff
@@ -66,7 +66,7 @@ env:
     - BUILD_CONAN_SERVER_IMAGE=1
 
 install:
-  - pip install conan
+  - pip install conan humanfriendly
```
uilianries: What humanfriendly does?

SSE4: outputs size as human readable value instead of big number, e.g. 10 Mb, 5 Gb, etc.

clang_3.8/Dockerfile
SSE4: okay

SSE4: tried to use `PYTHONDONTWRITEBYTECODE` (https://docs.python.org/3/using/cmdline.html), but `__pycache__ ` is still created :( will add comment.

PR #62, opened 2018-10-11
Add foreign architectures for package manager

ehargitt:
This is my first PR ever, so forgive me if I screw it up 😬. I added the foreign architecture just for the `gcc49-armv7hf` image for now so you could see what that would look like. I also added a statement to the `test.sh` script, but it looks like that doesn't get run for now anyway. The global variables in the `.travis.yaml` file were overriding the variables for the build plan in my Travis setup and causing the builds to fail. Consider removing them so others can fork and test in their own Travis builds. Anyway, let me know what you think and I could help do this for the other images.

PR #101, opened 2019-02-12
Add foreign architectures for package manager

ericLemanissier:
This is a rebased+squashed version of @ehargitt's https://github.com/conan-io/conan-docker-tools/pull/62

### DIFF
--- a/gcc_5-armv7hf/Dockerfile
+++ b/gcc_5-armv7hf/Dockerfile
@@ -13,6 +13,19 @@ ENV CC=arm-linux-gnueabihf-gcc-5 \
     LD=arm-linux-gnueabihf-ld \
     FC=arm-linux-gnueabihf-gfortran-5
 
+RUN sudo dpkg --add-architecture armhf
+RUN sudo sed -i 's/deb\s/deb \[arch=amd64,i386\] /' /etc/apt/sources.list
+
+USER root
+RUN PREFIX="deb [arch=armhf] http://ports.ubuntu.com/ubuntu-ports/" \
+  && SUFFIX="main restricted universe multiverse" \
+  && DEST="/etc/apt/sources.list.d/armhf.list" \
+  && echo "${PREFIX} xenial ${SUFFIX}" >> ${DEST} \
+  && echo "${PREFIX} xenial-updates ${SUFFIX}" >> ${DEST} \
+  && echo "${PREFIX} xenial-security ${SUFFIX}" >> ${DEST} \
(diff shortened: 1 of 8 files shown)


ericLemanissier:
I tested these images to build Qt for armv7hf, and it is able to install system packages via SystemPackageTool just flawlessly.

lasote:
Please @uilianries take a look

gcc_5-armv7hf/Dockerfile
```diff
@@ -13,7 +13,16 @@ ENV CC=arm-linux-gnueabihf-gcc-5 \
     LD=arm-linux-gnueabihf-ld \
     FC=arm-linux-gnueabihf-gfortran-5
 
-RUN sudo apt-get -qq update \
+RUN sudo dpkg --add-architecture armhf \
+    && sudo sed -i 's/deb\s/deb \[arch=amd64,i386\] /' /etc/apt/sources.list \
+    && PREFIX="deb [arch=armhf] http://ports.ubuntu.com/ubuntu-ports/" \
+    && SUFFIX="main restricted universe multiverse" \
+    && DEST="/etc/apt/sources.list.d/armhf.list" \
+    && echo "${PREFIX} xenial ${SUFFIX}" | sudo tee -a ${DEST} > /dev/null \
```
uilianries: I dont like to use pipe and echo on recipes, there is a recommendation about how to pipe on [Dockerfile Best Practices](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#using-pipes). However, I would recommend use only `printf` instead:

    printf "${PREFIX} xenial ${SUFFIX}" > ${DEST}

ericLemanissier: The problem is that sudo does not apply to stdout redirections

uilianries: ops! I forgot about this! :man_facepalming:
Could you please update to use `printf` instead of `echo`?
Also, you need to add `set -o pipefail` as recommended by Docker community.

Or, you can provide the expected file in the folder and just copy when building the image. I think is more readable for humans.

uilianries: I made a new commit with the proposed changes

ericLemanissier:
Indeed, it is much cleaner like this. thanks !

uilianries:
Conan server is broken, but is for other reason. I should submit a fix in another PR. I think this PR is good to go.

uilianries:
@lasote Any objection before to merge it?

PR #102, opened 2019-02-25
Mingw cross

theirix:
Add an image with `g++-mingw-w64-x86-64` to cross-build from Linux to Windows. Influenced by a useful https://github.com/bincrafters/docker-mingw-gcc7 image. This image is tested with building `zlib/1.2.11@conan/stable` and `libcurl/7.61.1@bincrafters/testing`.

Be aware that it does not use native gcc8 though it is based on `conanio/conangcc8`. The reason is to reuse the base image which is based on a newer `ubuntu:bionic` with g++-mingw-w64-x86-64 of version 7.


uilianries:
Since it is not listed on Travis file, it won't be built by our CI. Also, the scripts test.sh and build.sh are used only for manual builds, but would say they could be removed, since build.py is the only script used to build and test on CI jobs.

theirix:
@uilianries , thanks, I see now. I thought the compose file was a source of truth. Added a new
`- GCC_VERSIONS="7" DOCKER_ARCHS="x86_64" DOCKER_DISTRO="mingw"`
entry, hope it'll work. Shell scripts in each subdirectory are not so useful because they use `sudo docker`.
Can you please cancel a previous build?

SSE4:
@lasote would be nice if we could merge this one

Issue #105, opened 2019-02-26
gcc7-centos does not support system_requires: sudo -A fails


madebr:
Installing system requirements using the `tools.SystemPackageTool` fails in the `gcc7-centos` docker
The error is:
```
Running: sudo -A yum update -y
/var/tmp/sclH4Ekle: line 8: -A: command not found
```
travis: https://travis-ci.com/madebr/conan-ps2_toolchain_installer/jobs/180446149#L596

This can be reproduced in interactive mode:
```
$ docker run -t -i --rm -v /tmp:/home/conan/project:Z conanio/gcc7-centos6 /bin/bash
[conan@e5b2b8cb2901 ~]$ sudo -A ls
/var/tmp/sclvoMvA6: line 8: -A: command not found
```

uilianries:
that's because centos6 images are using ancient versions, including sudo. We will need to update sudo package or make some distinction for `SystemPackageTool`

madebr:
This is probably this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1319936

```
$ docker run -t -i --rm -v /tmp:/home/conan/project:Z conanio/gcc7-centos6 /bin/bash
[conan@034a47035262 ~]$ sudo -A ls
/var/tmp/sclcdiq0m: line 8: -A: command not found
[conan@034a47035262 ~]$ which sudo
/opt/rh/devtoolset-7/root/usr/bin/sudo
[conan@034a47035262 ~]$ /usr/bin/sudo -A ls
project
[conan@034a47035262 ~]$ 
```

I can fix it by replacing it with a symbolic link:
```
$ sudo rm -rf /opt/rh/devtoolset-7/root/usr/bin/sudo
$ /usr/bin/sudo ln -s /usr/bin/sudo /opt/rh/devtoolset-7/root/usr/bin/sudo
```

Or replacing the sudo script at `/opt/rh/devtoolset-7/root/usr/bin/sudo`
with the script at the bottom of the redhat issue tracker.


Afterwards, sudo works fine:
```
$ sudo -A ls
project
```

PR #106, opened 2019-02-28
Fix sudo script in gcc_7-centos6 and gcc_7-centos6-x86

madebr:
Fixes conan-io/conan-docker-tools#105

Tested on my local machine:
```
$ docker build --no-cache  -t conanio/gcc7-centos6_myversion .
$ docker run -t -i --rm -v /tmp:/home/conan/project:Z conanio/gcc7-centos6_myversion /bin/bash
[conan@48c653b979fa ~]$ sudo -A ls
project
```


uilianries:
Why not download a newer version and build it?
http://www.linuxfromscratch.org/blfs/view/8.2/postlfs/sudo.html

madebr:
Building a newer sudo will not fix this problem as the wrapper script is wrong.

uilianries:
Oh, now I got, the problem not `sudo` but its wrapper.

Issue #104, opened 2019-02-26
images to support cross-compilation for android


CAMOBAP795:
### Intro

Recently I pushed several PRs to fix/support https://github.com/conan-community/conan-opencv/pull/12 cross-compilation for Android

I did all test locally and then push it to review, before merge CI must be passed, for now CI covers linux & windows compilation only

### Suggestion

I would like to cover cross-compilation for Android too

### What I have tried

I have tried to run on https://github.com/dockcross/dockcross/tree/master/android-arm but it fails with [`/bin/sh: 1: cd: can't cd to project`](https://travis-ci.org/CAMOBAP795/conan-log4cplus/builds/497867759) error

Initially I thought that problem that user `conan` don't eixists in `dockcross/android-arm`, so I changed `CONAN_DOCKER_HOME` to `/root` but it doesn't help, issue remains the same

Full evn list: `ANDROID_NDK_HOME=/build/android-ndk-r16b CONAN_CLANG_VERSIONS=6.0 CONAN_DOCKER_IMAGE=dockcross/android-arm CONAN_DOCKER_IMAGE_SKIP_UPDATE=True CONAN_DOCKER_HOME=/root CONAN_ARCHS=armv7hf CONAN_BUILD_TYPES=Debug`

I will continue debugging locally

madebr:
I just had a similar issue and that issue was **selinux**.

`build.py` failed with the last line looking like: 
```
Exception: Error building: sudo -E docker run --rm -v "/home/maarten/programming/conan-ps2_toolchain_installer:/home/conan/project" -e CONAN_USERNAME="madebr" -e CONAN_DOCKER_IMAGE="conanio/gcc7" -e CONAN_CHANNEL="testing" -e CONAN_LOGIN_USERNAME="madebr" -e CONAN_REFERENCE="ps2_toolchain_installer/20190225@madebr/testing" -e CPT_PROFILE="@@include(default)@@@@[settings]@@arch_build=x86_64@@os_build=Linux@@[options]@@@@[env]@@@@[build_requires]@@@@" -e CONAN_TEMP_TEST_FOLDER="1" -e CPT_UPLOAD_RETRY="3"  conanio/gcc7 /bin/sh -c " cd project &&  run_create_in_docker "
```

When I added the `:Z` option to the bind mount, it worked on my system.

So I changed the command to
```
[...]

PR #107, opened 2019-02-28
add Android docker images

SSE4:
closes: #104

architectures:
- x86_64
- x86
- armv7
- armv8

tested with:

- zlib
- bzip2
- gtest
- lz4
- boost
- log4cplus
- OpenSSL

FYI there is a small CMake wrapper script to workaround some conan issues:
- https://github.com/conan-io/conan/issues/2402
- https://github.com/conan-io/conan/issues/4537
- https://github.com/conan-io/conan/issues/4629

/cc @CAMOBAP795 @madebr @theodelrieu @uilianries


uilianries:
@SSE4 Great job! Don't forget:

- Update travis to build the new docker image
- Update README with the new env vars DOCKER_CROSS and DOCKER_CACHE

android-clang_8-armv7/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    ADDR2LINE=$STANDALONE_TOOLCHAIN/bin/arm-linux-androideabi-addr2line \
+    NM=$STANDALONE_TOOLCHAIN/bin/arm-linux-androideabi-nm \
+    OBJCOPY=$STANDALONE_TOOLCHAIN/bin/arm-linux-androideabi-objcopy \
+    OBJDUMP=$STANDALONE_TOOLCHAIN/bin/arm-linux-androideabi-objdump \
+    READELF=$STANDALONE_TOOLCHAIN/bin/arm-linux-androideabi-readelf
+
+RUN conan profile update settings.arch=armv7 default \
+    && conan profile update settings.os=Android default \
+    && conan profile update settings.os.api_level=16 default \
+    && conan profile update settings.compiler.libcxx=libc++ default
```
CAMOBAP795: Should it be `libstdc++` https://github.com/conan-io/conan/issues/2856 ?

android-clang_8/Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM conanio/gcc8
```
CAMOBAP795: As I understand there is nothing special about `conanio/gcc8` we use it just to get `USER conan`, right?

CAMOBAP795:
@SSE4 Awesome, thank you!

android-clang_8-armv7/Dockerfile
SSE4: nope, with `libstdc++` it doesn't find <limits>. NDK r19b only supports libc++
https://developer.android.com/ndk/downloads/revision_history.html
`gnustl, gabi++, and stlport have been removed.` (from r18b)

android-clang_8/Dockerfile
SSE4: we may use any base image. we need things like CMake, make, autotools, nasm, etc., which are available in the base image.

SSE4:
okay, I've managed to solve OpenSSL build error. it used to default to API 24, so it needs to be passed `-D__ANDROID_API__=21` to build properly (see [NOTES.ANDROID](https://github.com/openssl/openssl/blob/master/NOTES.ANDROID)). unfortunately, it doesn't read `CFLAGS`, so I had to apply simple patch:
```
diff --git a/conanfile.py b/conanfile.py
index e12010d..ac1f660 100644
--- a/conanfile.py
+++ b/conanfile.py
@@ -140,6 +142,9 @@ class OpenSSLConan(ConanFile):
             extra_flags = "--debug" if self.settings.build_type == "Debug" else "--release"
             extra_flags += " no-shared" if not self.options.shared else " shared"

+        if "CFLAGS" in os.environ:
+            extra_flags += " " + os.environ["CFLAGS"]
+
         extra_flags += self._get_config_options_string()
         return extra_flags
```

CAMOBAP795:
@SSE4 How do you think in future make sense to move `
-D__ANDROID_API__=self.settings.os.api_level` this to `conan` (https://github.com/conan-io/conan/blob/develop/conans/client/build/cmake_flags.py) side?

SSE4:
@CAMOBAP795 for CMake I believe toolchain file sets everything needed
but for generic workflows, it could be applied here https://github.com/conan-io/conan/blob/develop/conans/client/build/compiler_flags.py
(to ensure not just CMake has required flags, but also Meson, AutoTools, etc.)

CAMOBAP795:
@SSE4 make sense. 

From other side according to `/opt/android-ndk-toolchains/arm-16/bin/arm-linux-androideabi-clang -E -dM - < /dev/null`, `__ANDROID_API__` should be defined by default (or I have lack of understanding how it works)

SSE4:
it should be, but OpenSSL tries to be smart, so it provides disservice and defines `__ANDROID_API__` if you didn't define it on command line explicitly...

SSE4:
@lasote conan_server image fails https://travis-ci.org/conan-io/conan-docker-tools/jobs/500249541:
```
unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
```
that doesn't seem to be related

uilianries:
Ops! Some packages need to be built, but there is no compiler installed:

```
Building wheels for collected packages: PyYAML, patch, future, typed-ast, lazy-object-proxy, wrapt
  Building wheel for PyYAML (setup.py): started
  Building wheel for PyYAML (setup.py): finished with status 'done'
  Stored in directory: /root/.cache/pip/wheels/ad/da/0c/74eb680767247273e2cf2723482cb9c924fe70af57c334513f
  Building wheel for patch (setup.py): started
  Building wheel for patch (setup.py): finished with status 'done'
  Stored in directory: /root/.cache/pip/wheels/42/cb/7a/9a7afb46eb44faf92333f3c6cd0e36865c39b84710e64d1d7d
  Building wheel for future (setup.py): started
  Building wheel for future (setup.py): finished with status 'done'
  Stored in directory: /root/.cache/pip/wheels/bf/c9/a3/c538d90ef17cf7823fa51fc701a7a7a910a80f6a405bf15b1a
  Building wheel for typed-ast (setup.py): started
  Building wheel for typed-ast (setup.py): finished with status 'error'
  Complete output from command /usr/local/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-ksej9fo6/typed-ast/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/pip-wheel-f9w_v3it --python-tag cp37:
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-3.7
  creating build/lib.linux-x86_64-3.7/typed_ast
  copying typed_ast/ast27.py -> build/lib.linux-x86_64-3.7/typed_ast
  copying typed_ast/conversions.py -> build/lib.linux-x86_64-3.7/typed_ast
  copying typed_ast/ast3.py -> build/lib.linux-x86_64-3.7/typed_ast
  copying typed_ast/__init__.py -> build/lib.linux-x86_64-3.7/typed_ast
  running build_ext
  building '_ast27' extension
  creating build/temp.linux-x86_64-3.7
  creating build/temp.linux-x86_64-3.7/ast27
  creating build/temp.linux-x86_64-3.7/ast27/Parser
  creating build/temp.linux-x86_64-3.7/ast27/Python
  creating build/temp.linux-x86_64-3.7/ast27/Custom
  gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -DTHREAD_STACK_SIZE=0x100000 -fPIC -Iast27/Include -I/usr/local/include/python3.7m -c ast27/Parser/acceler.c -o build/temp.linux-x86_64-3.7/ast27/Parser/acceler.o
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
```

SSE4:
@uilianries what should we do? add GCC to `conan_server` image? or change some dependencies?

uilianries:
that's weird, because it was not required on the last version. Since we need to build all those packages, we could add gcc, build, install and remove gcc, since we want a lightweight image.

SSE4:
@CAMOBAP795 FYI I gonna remove `__ANDROID_API__` from images, as it's only needed for OpenSSL, I believe it should be done in the recipe itself, this way it will work in case of any docker image or just standalone toolchain.

neobrain:
I ran across issues with these images trying to build sdl2 for armv8 (with the sdl2:shared=True option):

SDL links against dl, for which `find_library(X, dl)` returns `$SYSROOT/usr/lib/aarch64-linux-android/libdl.a`. That's just a stub implementation though, see https://android.googlesource.com/platform/bionic/+/master/libdl/libdl_static.cpp#21. So all uses of `dlopen` within libSDL.so return NULL.

Instead, SDL needs to be linked against the shared version of dl (i.e. libdl.so), which is located at `$SYSROOT/usr/lib/aarch64-linux/android/24/libdl.so/` (note the platform version being included in the latter path!).

I'm not sure if this is conan's fault here, or rather due to the way SDL looks for the dl library. Any suggestions?

neobrain:
~~I ran across issues with these images trying to build sdl2 for armv8 (with the sdl2:shared=True option):~~

~~SDL links against dl, for which `find_library(X, dl)` returns `$SYSROOT/usr/lib/aarch64-linux-android/libdl.a`. That's just a stub implementation though, see https://android.googlesource.com/platform/bionic/+/master/libdl/libdl_static.cpp#21. So all uses of `dlopen` within libSDL.so return NULL.~~

~~Instead, SDL needs to be linked against the shared version of dl (i.e. libdl.so), which is located at `$SYSROOT/usr/lib/aarch64-linux/android/24/libdl.so/` (note the platform version being included in the latter path!).~~

~~I'm not sure if this is conan's fault here, or rather due to the way SDL looks for the dl library. Any suggestions?~~

EDIT: Upon further investigation, this appears to be an issue with Android's CMake toolchain file. For further reference, see https://github.com/android-ndk/ndk/issues/929.

SSE4:
I gonna remove `ANDROID_NATIVE_API_LEVEL` from this PR as well
it's for compatibility only:
```
if(ANDROID_NATIVE_API_LEVEL AND NOT ANDROID_PLATFORM)
  if(ANDROID_NATIVE_API_LEVEL MATCHES "^android-[0-9]+$")
    set(ANDROID_PLATFORM ${ANDROID_NATIVE_API_LEVEL})
  elseif(ANDROID_NATIVE_API_LEVEL MATCHES "^[0-9]+$")
    set(ANDROID_PLATFORM android-${ANDROID_NATIVE_API_LEVEL})
  endif()
endif()
```
```
# Compatibility for read-only variables.
# Read-only variables for compatibility with the other toolchain file.
# We'll keep these around for the existing projects that still use them.
# TODO: All of the variables here have equivalents in our standard set of
# configurable variables, so we can remove these once most of our users migrate
# to those variables.
set(ANDROID_NATIVE_API_LEVEL ${ANDROID_PLATFORM_LEVEL})
```

SSE4:
updated to r19c

SSE4:
it doesn't build, because 3 images (x86, armv7 , armv8) depend on base one (x86_64), which is not exported into docker hub yet. I still think it's safe to merge.

PR #112, opened 2019-03-12
Add Clang 8 docker image

### LINKED ISSUE #97: Docker image for clang 8
Clang 8 is available: https://apt.llvm.org/bionic/dists/llvm-toolchain-bionic-8/

uilianries:
New Clang version is available on LLVM apt repo

closes #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ GCC>=5 is ABI compatible for minor versions. To solve multiple minors, there are
 | - [conanio/clang60: clang 6.0](https://hub.docker.com/r/conanio/clang60/)             | x86_64 |  Supported |
 | - [conanio/clang7-x86: clang 7](https://hub.docker.com/r/conanio/clang7-x86/)         | x86    |  Supported |
 | - [conanio/clang7: clang 7](https://hub.docker.com/r/conanio/clang7/)                 | x86_64 |  Supported |
+| - [conanio/clang8-x86: clang 8](https://hub.docker.com/r/conanio/clang8-x86/)         | x86    |  Supported |
+| - [conanio/clang8: clang 8](https://hub.docker.com/r/conanio/clang8/)                 | x86_64 |  Supported |
 
 
 #### Conan Server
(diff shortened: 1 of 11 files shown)


Croydon:
@lasote's argument for not upgrading distros was that we are using Ubuntu versions which did have the compilers versions "originally". With the same argument you could argument against this pull request. Maybe use https://packages.ubuntu.com/disco/clang-8?

We could also argue that a LTS release would be wiser, but than we should use ppa's for all Conan docker images. 🤔

uilianries:
It's not possible to use Disco because is under development and some packages are broken:

https://bugs.launchpad.net/ubuntu/+source/llvm-defaults/+bug/1819768

As the release will occur in April, I gonna update this PR and wait for a solution.

uilianries:
I'll drop x86 support for clang-8 until the package is not fixed

lasote:
Some other builds have fail. What is the status of this clang 8 support?

uilianries:
There is no status. I've opened the bug on lauchpad, but it's parked there. In fact, there are 3 bugs reported for the llvm-tools there, only one is being supported for now. I'm not sure if the llvm team will solve this.

About the error, it looks like some packages are outdated again. I'm going to push a new PR with a fix, because even 1.14.4 is being affected.

Croydon:
What exactly is the error/problem? What happens when you try to install/execute?

uilianries:
libc++:x86 fails to be installed on x86_64

SSE4:
there is an error:
```
usr/bin/ld: skipping incompatible /usr/lib/llvm-8/bin/../lib/libc++.so.1 when searching for libc++.so.1
/usr/bin/ld: skipping incompatible /usr/lib/x86_64-linux-gnu/libc++.so.1 when searching for libc++.so.1
/usr/bin/ld: cannot find libc++.so.1
/usr/bin/ld: skipping incompatible /usr/lib/llvm-8/bin/../lib/libc++abi.so when searching for -lc++abi
/usr/bin/ld: skipping incompatible /usr/lib/llvm-8/bin/../lib/libc++abi.a when searching for -lc++abi
/usr/bin/ld: skipping incompatible /usr/lib/x86_64-linux-gnu/libc++abi.so when searching for -lc++abi
/usr/bin/ld: cannot find -lc++abi
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
https://travis-ci.com/bincrafters/conan-gtest/jobs/198278773

PR #114, opened 2019-03-12
#81 Install autoconf-archive

### LINKED ISSUE #81: autoconf-archive might be useful
Hi, it might be useful to add autoconf-archive to the docker images, in order to handle configure.ac files that utilise AX_ macros from autoconf-archive. I've just struck this while packaging libgpiod. I'm happy to put together a PR if this would be of interest.

uilianries:
Install autoconf-archive. It's small only some KB

fixes #81 
/cc @mbeach


uilianries:
Thanks @madebr ! :+1:

PR #116, opened 2019-03-12
Fix Conan Server startup

uilianries:
related issue: https://github.com/conan-io/conan/issues/4723

### DIFF
--- a/conan_server/Dockerfile
+++ b/conan_server/Dockerfile
@@ -15,6 +15,7 @@ RUN apk add --no-cache gcc libc-dev \
     && pip3 install -r conans/requirements_server.txt \
     && pip3 install gunicorn \
     && apk del gcc libc-dev \
+    # Do not migrate anything. Only populate ~/.conan_server
     && PYTHONPATH=${PWD} python conans/conan_server.py --migrate
 
 EXPOSE 9300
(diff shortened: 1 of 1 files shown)


SSE4:
issue says `conan_server -migrate` may fail. why doesn't it happen here?
also, why do we need to migrate during the build?
I would expect to do this in entry point, at least.

conan_server/Dockerfile
```diff
@@ -13,8 +13,10 @@ RUN apk add --no-cache gcc libc-dev \
     && cd conan-${CONAN_VERSION} \
     && pip3 install -r conans/requirements.txt \
     && pip3 install -r conans/requirements_server.txt \
+    && pip3 install . \
```
lasote: Why this?

uilianries: the problem occurs because gnunicorn runs 4 workers at startup and it cause a conflict because all the workers try to create ~/.conan_server/data at same time.

I installed conan to be to execute conan_server --migrate and populate ~/.conan_server/data.

The point is that we need to fix the race condition when there is no data folder there. Any better idea?

SSE4: why not just create a folder then (`mkdir`)?

uilianries: because there are also `server.conf` and `version.txt` at same level.

uilianries: Now it won't install Conan, but --migrate is required just to populate the conan_server folder.

PR #123, opened 2019-03-27
Add Template for Github issues ***NO_CI***

uilianries:
- Add Contributing guide from CPT
- Add issues template from CPT
- Add issues labels from CPT

TIL: [How to skip CI on Azure DevOps](https://github.com/Microsoft/azure-pipelines-agent/issues/858#issuecomment-475768046)

EDIT: `***NO_CI***` does not work!

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,86 @@
+Contributing to Conan Package Tools
+===================================
+
+The following summarizes the process for contributing to the Conan Docker Tools project.
+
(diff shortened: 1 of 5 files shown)


uilianries:
Since this PR is only related to gihub management, we don't need to run the CI.

But since Azure doesn't follow other CI services, like Travis, Circle, Appveyor ... it uses `***NO_CI***` instead of `[skip ci]` on commit message. However, even including this special word, the CI build started on Azure.

PR #127, opened 2019-04-15
define a custom profile instead of default

### LINKED ISSUE #126: [bug] Android images detect Linux on default profile
### Environment Details
* Conan Docker Tools image: **conanio/xxx-android**
* Operating System: **Linux**
* Operation System Version: **Ubuntu 16.04**
* Python version (If you are building some image): **python 3.7.0**

### Steps to reproduce (Include if Applicable)

    docker run --rm -ti conanio/gcc8-android
    conan profile show default

/cc @ericLemanissier

ericLemanissier:
The default profile is always overwritten by conan package tools, 
so settings are now set in a custom profile: android

fixes #126

- [X] Refer to the issue that supports this Pull Request.
- [X] I've read the [Contributing guide](https://github.com/conan-io/conan-docker-tools/blob/master/.github/CONTRIBUTING.md).
- [X] I've followed the [PEP8](https://www.python.org/dev/peps/pep-0008) style guides for Python code.
- [X] I've followed the [Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) guides for Dockerfile.

### DIFF
--- a/android-clang_8-armv7/Dockerfile
+++ b/android-clang_8-armv7/Dockerfile
@@ -17,7 +17,7 @@ ENV ANDROID_ABI=armeabi-v7a \
     OBJDUMP=$STANDALONE_TOOLCHAIN/bin/arm-linux-androideabi-objdump \
     READELF=$STANDALONE_TOOLCHAIN/bin/arm-linux-androideabi-readelf
 
-RUN conan profile update settings.arch=armv7 default \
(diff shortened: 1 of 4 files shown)


ericLemanissier:
It seems this is not the proper solution. CPT does all the profile handling outside of docker, so any profile created inside the docker image is actually useless.

PR #133, opened 2019-05-13
Use sources.list instead of sed to update sources.list

uilianries:
sed works, but it is complicated, which makes it difficult to understand what has been modified.

Changelog: Fix: Use sources.list instead of sed to update sources.list

### DIFF
--- a/clang_3.9/sources.list
+++ b/clang_3.9/sources.list
@@ -1,0 +1,18 @@
+deb http://old-releases.ubuntu.com/ubuntu/ artful main restricted
+
+deb http://old-releases.ubuntu.com/ubuntu/ artful-updates main restricted
+
+deb http://old-releases.ubuntu.com/ubuntu/ artful universe
+deb-src http://old-releases.ubuntu.com/ubuntu/ artful universe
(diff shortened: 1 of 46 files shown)


clang_3.9/sources.list
```diff
@@ -0,0 +1,18 @@
+deb http://old-releases.ubuntu.com/ubuntu/ artful main restricted
```
SSE4: is this file different from `clang_3.9-x86/sources.list`?

uilianries: only a new line is different, but the content is the same. I checked with vimdiff

SSE4: okay, may we reduce copy-paste somehow? e.g. move file to some shared location, or just use it in base image once?

uilianries: I think the second options is better, since we can't use relative path when copying (docker feature).

uilianries: I cant use the second options ... docker-compose uses each image folder as a context, which means it won't the sources.list file. I could use symbolic link WDYT ?

SSE4: that's unfortunate, okay, let's leave it for now as is

PR #129, opened 2019-04-26
Add GCC 9

Croydon:
Changelog: Feature: Add GCC 9 containers in several flavors

- [x] I've read the [Contributing guide](https://github.com/conan-io/conan-docker-tools/blob/master/.github/CONTRIBUTING.md).
- [x] I've followed the [PEP8](https://www.python.org/dev/peps/pep-0008) style guides for Python code.
- [x] I've followed the [Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) guides for Dockerfile.

Bases on Ubuntu Disco

PR #134, opened 2019-05-14
One CI job, many docker images

### LINKED ISSUE #125: Use Docker multi stage to reduce the recipe number

Months ago @SSE4 commented that we could use [Docker multistage build](https://docs.docker.com/develop/develop-images/multistage-build/) to reduce the number of recipes, but we had some limitation related to Docker Compose version supported by Travis. Since we have migrated to Azure we could re-visit this idea and try to optimize our Docker recipes.

All extra arch images, like ARM, could be merged into the default Docker recipe.

Related issue #89

uilianries:
Hi!

Since we are having trouble with CDT to re-use basic images (e.g. conanio/gcc8), this PR tries to unify many CI jobs as possible.

Before this PR I tried 2 other alternatives, but without sucess:
- Using Azure stages to build first all base images, and a second stage to build specialized images. It didn't work because it's not possible retrieving a docker image from previous stage. The experiment can be verified [here](https://github.com/uilianries/conan-docker-tools/blob/feature/multiple-stage/azure-pipelines.yml).
- Docker [multistage build](https://docs.docker.com/develop/develop-images/multistage-build/) is an incredible feature, so why not creating an unique Docker recipe with many different images? So I've created a docker [recipe](https://gist.github.com/uilianries/d691e55065791ca3e87172195fe2243f) for GCC8, including x86_64, armv7, armv7hf and armv8. However is not possible skipping a specific target when building ... I found [build_kit](https://docs.docker.com/develop/develop-images/build_enhancements/) which offers some flexibility, but it's not supported by docker-compose.

So this PR basically build all those images as before, but now all specialized images are on same CI job and following a specific order. For example, to build GCC8, first we build x86_64, armv7 and so on. Thus we don't need to re-build same x86_64 as before. The same idea is applied for distros, where we only need x86_64 to build Jenkins' images.

**PROS**:
- Less building time
- More re-usability of base images (x86_64)

**CONS**:
- Populated log containing many builds on same place.  

fixes #125 

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,5 +310,4 @@ Upload related variables:
 - **DOCKER_PASSWORD**: Your Docker password to authenticate in Docker server
 - **DOCKER_UPLOAD**: If attributed to true, it will upload the generated docker image, positive words are accepted, e.g "True", "1", "Yes". Default "False"
 - **BUILD_CONAN_SERVER_IMAGE**: If attributest to true, it will build and upload an image with the conan_server
-- **BUILD_BASE_DISTRO**: Build base Docker image before building distro image. By default is True.
 - **DOCKER_UPLOAD_ONLY_WHEN_STABLE**: Only upload only when is master branch.
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,257 +15,129 @@ jobs:
       DOCKER_PASSWORD: $(DOCKER_PASSWORD)
   strategy:
     matrix:
-      Jenkins Slave GCC 4.6 x86_64:
-        GCC_VERSIONS: "4.6"
-        DOCKER_ARCHS: "x86_64"
-        DOCKER_DISTRO: "jnlp-slave"
-      Jenkins Slave GCC 4.8 x86_64:
-        GCC_VERSIONS: "4.8"
-        DOCKER_ARCHS: "x86_64"
-        DOCKER_DISTRO: "jnlp-slave"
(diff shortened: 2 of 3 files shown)


lasote:
> Less building time

What is the improvement?

uilianries:
> What is the improvement?

Instead of building the same base image (x86_64) for each specific image (armv7, armv7hf, armv8), we create a CI job with all those images, which includes x86_64 only once.

lasote:
I was asking about time improvements, 20%, 30%, 5%?

uilianries:
ouch ... 60% more time than before! :disappointed: 

The current CDT takes 1h50 to build all images, with this new change it will take 2h50.

I have have 2 hypotheses:
- Since the new jobs take a longer time to be built, we will wait more time.
- Some networking delay at that time increased build time.

I just started a new build to check the times again.

Croydon:
Where did you get the 2h50m from? https://dev.azure.com/conanio/conan-docker-tools/_build/results?buildId=233&view=results this here is showing 1h50m 🤔

uilianries:
This one:

https://dev.azure.com/uilianr/conan-devops/_build?definitionId=3

Croydon:
Seems to be a temporary spike, your last job was 1h8mins

Croydon:
@lasote Could you please review this? We need this pull request merged so that the CI for #129 can succeed (or any other merge request which is adding new containers)

PR #142, opened 2019-08-08
install Clang 8 from the http://apt.llvm.org/

SSE4:
/cc @uilianries 

fix Clang 8 image, I was able to run `conan create . bincrafters/testing -k -s arch=x86 -s compiler.libcxx=libc++` for `conan-libx265`.

### DIFF
--- a/clang_8/Dockerfile
+++ b/clang_8/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:disco
+FROM ubuntu:cosmic
 
 LABEL maintainer="Luis Martinez de Bartolome <luism@jfrog.com>"
 
@@ -17,7 +17,6 @@ RUN dpkg --add-architecture i386 \
        wget=1.* \
        git=1:2.* \
        g++-multilib=4:8.* \
-       clang-8=1:8* \
        make=4.* \
        libc6-dev-i386=2.* \
        libgmp-dev=2:6.* \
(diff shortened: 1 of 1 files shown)


uilianries:
Cosmic is under development, did you check if LLVM offers the same package for bionic ?

SSE4:
yes, it does: http://apt.llvm.org/

uilianries:
I remember that @Croydon commented something about using LLVM repo instead of Ubuntu. I think was something related to ABI compatibility.

Croydon:
It's probably better to not downgrade to distribution when it is also available for Disco

> I remember that @Croydon commented something about using LLVM repo instead of Ubuntu. I think was something related to ABI compatibility.

I guess I might have asked at some point if we can use LLVM's repos to install Clang, yes.

uilianries:
We could check if is compatible, using LLVM repo for Bionic. How? Linking a package which was built using our current clang8 image, with a new project that uses clang from LLVM repo.

Croydon:
The Clang 8 image is currently using Ubuntu Disco (19.04), so it is a stable release and LLVM is offering packages for Disco. Why should we use Bionic packages?

SSE4:
I personally don't care much if it's cosmic/disco/bionic/whatever. I am more interested in getting working Clang 8 image first. so, I can use whatever you prefer.

Croydon:
Why closing? Shouldn't we still fix this?

uilianries:
well, we will need to use another distro, as disco doesn't provide x86 support for libc++. However, using another distro can implicate to glibc compatibility. 

The same problem occurred for clang 9. Although, we dropped x86 support, so I'm sure if we should invest more time on this very specific issue. Do you have some suggestion?

Croydon:
Okay, I'm fine with this. However, this kinda raises the question if we should continue to add 32bit containers for new compiler versions. Probably not.

uilianries:
Sure, we can drop. If someone asks for such support, we can add again, since it's not complicated.

PR #144, opened 2019-09-02
set CONAN_CLANG/GCC_VERSIONS env var

ericLemanissier:
This simplifies the use of conan package tools in CI: with this change one can set a matrix job setting only the docker image to use, and CPT automatically picks up the proper GCC or clang version, based on the environment variables set by the image

### DIFF
--- a/clang_9-x86/Dockerfile
+++ b/clang_9-x86/Dockerfile
@@ -11,6 +11,8 @@ ENV LLVM_VERSION=9.0 \
     CONAN_ENV_ARCH=x86 \
     PYENV_ROOT=/opt/pyenv \
     PATH=/opt/pyenv/shims:${PATH}
+	
+COPY sources.list /etc/apt/sources.list
 
 RUN apt-get -qq update \
     && apt-get -qq install -y --no-install-recommends \
--- a/clang_9-x86/sources.list
+++ b/clang_9-x86/sources.list
@@ -1,0 +1,17 @@
+deb http://old-releases.ubuntu.com/ubuntu/ eoan main restricted
+deb http://old-releases.ubuntu.com/ubuntu/ eoan-updates main restricted
+
+deb http://old-releases.ubuntu.com/ubuntu/ eoan universe
+deb-src http://old-releases.ubuntu.com/ubuntu/ eoan universe
+deb http://old-releases.ubuntu.com/ubuntu/ eoan-updates universe
+deb-src http://old-releases.ubuntu.com/ubuntu/ eoan-updates universe
+
+deb http://old-releases.ubuntu.com/ubuntu/ eoan multiverse
+deb http://old-releases.ubuntu.com/ubuntu/ eoan-updates multiverse
+
+deb http://old-releases.ubuntu.com/ubuntu/ eoan-backports main restricted universe multiverse
+
+deb http://old-releases.ubuntu.com/ubuntu/ eoan-security main restricted
+deb http://old-releases.ubuntu.com/ubuntu/ eoan-security universe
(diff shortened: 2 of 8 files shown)


uilianries:
We don't install CPT directly in our docker images, also CPT is who controls the docker image, including the compiler version:

export CONAN_GCC_VERSIONS=8
python build.py -> (read env vars) CPT -> (pass env vars) Docker -> CPT (in docker)

But since it does not affect Conan directly, I see no problem.

android-clang_8/Dockerfile
```diff
@@ -32,7 +32,9 @@ ENV ANDROID_NDK=$ANDROID_NDK \
     CMAKE_FIND_ROOT_PATH_MODE_LIBRARY=BOTH \
     CMAKE_FIND_ROOT_PATH_MODE_INCLUDE=BOTH \
     CMAKE_FIND_ROOT_PATH_MODE_PACKAGE=BOTH \
-    PATH=$PATH:$STANDALONE_TOOLCHAIN/bin
+    PATH=$PATH:$STANDALONE_TOOLCHAIN/bin \
+    CONAN_CLANG_VERSIONS=8 \
+    CONAN_GCC_VERSIONS=
```
lasote: This is to unset the value or...?

ericLemanissier: Yes, because it was set by base image

ericLemanissier:
The drawback of letting CPT handle docker containers is that it spawns 1 fresh container to build each binary configuration, even if several share the same compiler version and arch (which means they actually use the same docker image). Handling docker containers at CI level allows to reduce build times by installing system_requirements once for all configurations using same compiler version and arch(usually 4 configs: release/debug*shared/static). Also, package required are downloaded once for all configuration using same compiler version, arch, and build_type(usually 2 configs:shared/static).
Having this change simplifies this use case, because you only have to specify the docker image at CI level, and CPT automatically knows the compiler version used. And now that I think about it, `CONAN_ARCHS` env variable could also be set inside docker images. I'll add it too if you are ok with the change.

ericLemanissier:
Well it's still nice to have the same interface `python build.py` regardless of who manages docker containers. also, having CPT handle the build pages, CONAN_LOGIN_USERNAME, CONAN_PASSWORD and CONAN_BUILD_REQUIRES for example is still valuable.

clang_3.8/Dockerfile
```diff
@@ -9,7 +9,8 @@ ENV LLVM_VERSION=3.8 \
     CMAKE_CXX_COMPILER=clang++ \
     PYENV_ROOT=/opt/pyenv \
     PATH=/opt/pyenv/shims:${PATH} \
-    CONAN_CLANG_VERSIONS=3.8
+    CONAN_CLANG_VERSIONS=3.8 \
+    CONAN_ARCHS=x86_64
```
uilianries: Why? We can cross-compile to x86 using this image. Same for other images on this commit.

ericLemanissier: I can put `CONAN_ARCHS=x86_64,x86` if your prefer. Environment variables can be set from CI anyway, It's just a matter of choosing the most common use case. It looked more consistent to have only x86_64, with conan-center-index supporting x86_64 only, but I have no strong feeling about this.

uilianries: Yeah, but Conan Center Index is not the only one who uses these docker images. We are using for Bincrafters too. Of course, we dropped x86 support, I would say someone will complain about this. Please, don't touch CONAN_ARCHS

android-clang_8/Dockerfile
ericLemanissier: @lasote are you ok with this ?

Issue #145, opened 2019-09-15
alias for clang7 image


SSE4:
at some point, we may decide to make an additional alias for the `clang7` image, e.g. `clang70`.
/cc @uilianries @Hipony @Minimonium

Croydon:
Is there a real docker alias function or would we need to build and upload it a second time?

Minimonium:
You'd need to run the `docker tag`. I'm not sure how the build script should handle it in a not that hacky way though.

Example:
```
$ sudo docker tag conanio/clang7 conanio/clang70
$ sudo docker image ls
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
conanio/clang7      latest              e96ae83f7f28        2 days ago          856MB
conanio/clang70     latest              e96ae83f7f28        2 days ago          856MB
```

Croydon:
Done!

PR #147, opened 2019-09-16
Create alias for clang 7

uilianries:
Since Clang 7 is available as clang 7.0 in Conan settings and LLVM just released Clang 7.1, I think this alias can help to build with CPT:

    CONAN_CLANG_VERSIONS=7.0 CONAN_DOCKER_IMAGE=clang70

Original discussion: https://cpplang.slack.com/archives/C41CWV9HA/p1568472642006400

/cc @Minimonium @ericLemanissier 

Changelog: Feature: Create alias for clang7 as clang70

### DIFF
--- a/build.py
+++ b/build.py
@@ -319,12 +319,24 @@ class ConanDockerTools(object):
         logging.info("Upload Docker image %s" % self.tagged_image_name)
         subprocess.check_call("docker push %s" % self.tagged_image_name, shell=True)
 
+        if self.service == "clang7":
+            logging.info("Clang 7 will upload the alias Clang 7.0")
+            subprocess.check_call("docker push %s" %
+                self.tagged_image_name.replace("clang7", "clang70"), shell=True)
+
     def tag(self):
         """Apply Docker tag name
         """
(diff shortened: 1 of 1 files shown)


Minimonium:
Closes #145

ericLemanissier:
I'm not a docker expert, but it looks like this PR does not properly create the `latest` tag for conanio/clang70 image: https://dev.azure.com/bincrafters/packages/_build/results?buildId=612
there is no `latest` tag here https://hub.docker.com/r/conanio/clang70/tags?page=1 contrary to `https://hub.docker.com/r/conanio/clang7/tags?page=1`

lasote:
@uilianries please take a look

lasote:
@ericLemanissier could you better open an issue? to keep better track

uilianries:
It's the same, you can check the digest, but latest is not provide only. We should improve that.

Croydon:
We need the latest tag in order to being able to pull that container without explicitly specifying a tag (and also obviously to pull always the latest version)

uilianries:
I agree! I going to provide a fix soon

PR #153, opened 2019-10-18
Update python version to 3.8.0

### LINKED ISSUE #150: Update Python version to 3.8
[Python 3.8](https://docs.python.org/3/whatsnew/3.8.html) is available for [download](https://www.python.org/downloads/release/python-380/)

uilianries:
Update python version used by Docker images
closes #150 
Changelog: Feature: Update python version to 3.8.0

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,7 +4,7 @@
 * Conan Docker Tools image: **conanio/gcc7:1.13.1**
 * Operating System: **Linux**
 * Operation System Version: **Ubuntu 16.04**
-* Python version (If you are building some image): **python 3.7.0**
+* Python version (If you are building some image): **python 3.8.0**
 
 ### Steps to reproduce (Include if Applicable)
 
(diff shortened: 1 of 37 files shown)


danimtb:
Conan does not support 3.8 as there are some issues reported. So we cannot update the images yet

uilianries:
Conan supports Python 3.8 since https://github.com/conan-io/conan/pull/6439

uilianries:
@danimtb I think we are good now.

danimtb:
This PR also unifies the Python version used in all images, right?

uilianries:
@danimtb Yes, all images will run Python 3.8.5

memsharded:
> @danimtb Yes, all images will run Python 3.8.5

Why? Wouldn't it be a bit more conservative to run something more backwards compatible. Running with latest Python, users can use some newer Python features that will only work in Python 3.8, like the walrus operator (then breaking ConanCenter for everyone using Python<3.8), while doing something in Python 3.6 that breaks in Python 3.8 is way more unlikely.

danimtb:
I see we were already using Python 3.7.5 in almost all images so for me it has more value the unification of the version than the update. I don't know what has been the policy updating python in conan-docker-tools over the time, but I tend to think the same way as @memsharded, is the update really needed? do we have users asking for this change?

uilianries:
Not really required for now, closing

PR #173, opened 2020-01-15
Install 'python3.8-distutils' too

jgsogo:
We need `virtualenv --python "/usr/bin/python3.8"` to work inside the image out of the box

Tip from here https://github.com/deadsnakes/issues/issues/82#issuecomment-478288319


jgsogo:
@uilianries , feel free to merge if it doesn't fail   🐳

PR #177, opened 2020-01-23
Use xenial for GCC4.9

uilianries:
Since Conan Center Index is using GCC5 for building installer packages, most of them won't work with package with are built with GCC 4.9 due the Ubuntu version used in the docker image which is Trusty (14.04). 

Updating the base images to xenial (16.04) will solve that problem, however, as side effect, some users that are still running trusty will find the same compatibility error.

IMO I don't see people asking for GCC 4.x anymore, so I presume any side effect will be minimal or even for companies which prefer building packages from sources.

An example of this problem: https://github.com/bincrafters/conan-glib/pull/13/checks?check_run_id=404903770

The project glib requires `meson/0.53.0`, which was built on xenial by gcc5. As it's an installer package, compiler is not part of its ID, thus the result is an error due the glibc version.

Changelog: Fix: 4.9 now is compatible with GCC 5


ericLemanissier:
Isn't this the kind of change which would break the ABI between binaries created with old image (wily) and new image (xenial) ?

uilianries:
@ericLemanissier Yes, it will break. However, we can re-generate all GCC 4.9 packages. Also, do you remember any case where people asked for GCC 4.x ? The problem here is that CCI will break some packages due glic version in Xenial.

ericLemanissier:
I personally don't care about GCC 4.x, and I don't recall specific requests about it
I think all the conan docker tools image should be based on the same distribution as CCI for the same compiler version, otherwise it makes it impossible to combine CCI binaries with binaries compiled on conan docker tools images.

uilianries:
> I think all the conan docker tools image should be based on the same distribution as CCI for the same compiler version, otherwise it makes it impossible to combine CCI binaries with binaries compiled on conan docker tools images.

I agree. CCI is running these docker images, so we can "fix" that problem. I would like to use the same distro for any compiler, but is more complicated, since we don't have all them available for a same distro version.

Croydon:
I'm guessing this could be fine for now

For the long-term we need a better strategy how to handle glibc compatibility. Relying on end-of-life Ubuntu versions can't be the answer

Croydon:
Friendly ping @jgsogo @danimtb 

Many CI pipelines for GCC 4.9 are broken since we started moving installer packages to CCI. Unbreaking them would be nice 😄

ericriff:
Is this going to get merged?

uilianries:
@ericriff No. We have a new propose: #204

PR #184, opened 2020-03-06
Install CMake via pip

Croydon:
Changelog: Feature: Install CMake via pip

This simplifies the installation of CMake a lot. In fact, different containers used different line of codes and weren't always up to date (like excluding paths which don't exist in the current used CMake version).

Bincrafters' CI uses CMake via pip for a long time and personally that is how I'm managing my CMake installations as well on Fedora and Windows.

### DIFF
--- a/clang_3.8/Dockerfile
+++ b/clang_3.8/Dockerfile
@@ -71,8 +71,8 @@ RUN dpkg --add-architecture i386 \
     && update-alternatives --install /usr/bin/pyenv pyenv /opt/pyenv/bin/pyenv 100 \
     && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.7.5 \
     && pyenv global 3.7.5 \
-    && pip install -q --upgrade --no-cache-dir pip \
-    && pip install -q --no-cache-dir conan conan-package-tools scikit-build cmake==3.16.* \
+    && pip install -q --upgrade --no-cache-dir pip scikit-build \
+    && pip install -q --no-cache-dir conan conan-package-tools cmake==3.16.* \
     && chown -R conan:1001 /opt/pyenv \
     # remove all __pycache__ directories created by pyenv
     && find /opt/pyenv -iname __pycache__ -print0 | xargs -0 rm -rf \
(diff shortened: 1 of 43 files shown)


uilianries:
What!? Since when cmake is python?

Croydon:
It is not, but you can ship native executables via pip. Actually, a lot of packages on pip are doing this which rely on native Python modules (e.g. a lot of machine learning packages are doing this for additional speed)

Croydon:
But it looks like they not providing packages for 32 bit. So maybe we should do this only for new containers (I'm on Clang 10 / GCC 10 for x86_64 right now)

Croydon:
> ERROR: conan-package-tools 0.31.2 has requirement six<1.13.0,>=1.10.0, but you'll have six 1.14.0 which is incompatible.


Not sure what to do about this

uilianries:
> So maybe we should do this only for new containers 

We will continue providing x86 images, thus we can apply it only for x86_64 images. Even Cmake doesn't distribute pre-built binaries for x86.

Croydon:
On the x86 images we build CMake ourselves. I believe that the pip distribution can do it too. So lets see first if it works

But this is completely blocked by conan-io/conan-package-tools#477, due to restrictions on the `six` package version

uilianries:
CPT 0.32.0 should be release on this week

clang_3.8/Dockerfile
```diff
@@ -79,8 +71,8 @@ RUN dpkg --add-architecture i386 \
     && update-alternatives --install /usr/bin/pyenv pyenv /opt/pyenv/bin/pyenv 100 \
     && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.7.5 \
     && pyenv global 3.7.5 \
-    && pip install -q --upgrade --no-cache-dir pip \
-    && pip install -q --no-cache-dir conan conan-package-tools \
+    && pip install -q --upgrade --no-cache-dir pip scikit-build \
+    && pip install -q --no-cache-dir conan conan-package-tools cmake==3.16.* \
```
uilianries:
```suggestion
    && pip install -q --no-cache-dir conan conan-package-tools cmake==3.16.4 \
```

You need to use a very specific version, because we use a script to update all images. Also, it keeps a more controlled scenario.

Croydon:
Closing in favour of https://github.com/conan-io/conan-docker-tools/pull/241

PR #193, opened 2020-04-17
Add Clang 10 image

Croydon:
Changelog: Feature: Add Clang 10 container image

__________________________________

I decided to go with the upstream LLVM APT repository this time, as they offer builds for older LTS versions of Ubuntu too.

Ubuntu itself is currently only offering Clang 10 binaries for Ubuntu 20.04. Using this would mean a brand new glibc version which causes that the built binaries are compatible with just very few distribution versions out there....

Ubuntu Bionic seems currently to be a fair compromise between not too old and not too new.

I also re-structured the Dockerfile a little bit differently compared to previous Clang containers, so that it is more clear what packages are installed why.

LLVM, Python and CMake versions are now also specified via environment variables at the beginning of the Dockerfile. Making it easier to read/maintain those.

Additional things that I did not want to split up in other pull requests:
  * Using recipes from CCI instead the old ones for testing
  * Update README examples with more up-to-date compiler versions

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -719,6 +719,15 @@ services:
         image: "${DOCKER_USERNAME}/clang9-jnlp-slave-x86:${DOCKER_BUILD_TAG}"
         container_name: clang9-jnlp-slave-x86
         tty: true
+    clang10-jnlp-slave:
+        build:
+            context: jenkins-jnlp-slave
+            dockerfile: Dockerfile
+            args:
+                SOURCE_CONANIO_IMAGE: conanio/clang10
+        image: "${DOCKER_USERNAME}/clang10-jnlp-slave:${DOCKER_BUILD_TAG}"
+        container_name: clang10-jnlp-slave
+        tty: true
     conantests:
         build:
             context: conan_tests
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -400,6 +400,13 @@ services:
         image: "${DOCKER_USERNAME}/clang9-x86:${DOCKER_BUILD_TAG}"
         container_name: clang9-x86
         tty: true
+    clang10:
+        build:
+            context: clang_10
+            dockerfile: Dockerfile
+        image: "${DOCKER_USERNAME}/clang10:${DOCKER_BUILD_TAG}"
+        container_name: clang10
+        tty: true
     msvc14:
         build:
             context: msvc_14
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ You can use these images directly in your project or with the [conan-package-too
 > :warning: **Warning:**
 The images listed below are intended for **generating open-source library packages** and we cannot guarantee any kind of stability. We strongly recommend using your own generated images for production environments taking the dockerfiles in this repository as a reference.
 
-The images are uploaded to Dockerhub:
+The images are uploaded to Docker Hub:
 
 #### GCC
 | Version                                                                                       | Arch    |  Status, Life cycle  |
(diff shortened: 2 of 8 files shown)


uilianries:
please, create separated PRs for Cmake and clang 10

Croydon:
> /bin/sh: 1: cannot open clang_10-jnlp-slave/Dockerfile: No such file

I remember that this was already last time a problem when I suggested to add a new container via PR, but I forgot the solution to it. 🤔

docker-compose.yml
```diff
@@ -400,6 +400,13 @@ services:
         image: "${DOCKER_USERNAME}/clang9-x86:${DOCKER_BUILD_TAG}"
         container_name: clang9-x86
         tty: true
+    clang10:
+        build:
+            context: clang_10
+            dockerfile: Dockerfile
+        image: "${DOCKER_USERNAME}/clang10:${DOCKER_BUILD_TAG}"
+        container_name: clang10
+        tty: true
```
uilianries, line 409:
```suggestion
        tty: true
    clang10-x86:
        build:
            context: clang_10-x86
            dockerfile: Dockerfile
        image: "${DOCKER_USERNAME}/clang10-x86:${DOCKER_BUILD_TAG}"
        container_name: clang10-x86
        tty: true
    clang10-jnlp-slave:
        build:
            context: jenkins-jnlp-slave
            dockerfile: Dockerfile
            args:
                SOURCE_CONANIO_IMAGE: conanio/clang10
        image: "${DOCKER_USERNAME}/clang10-jnlp-slave:${DOCKER_BUILD_TAG}"
        container_name: clang10-jnlp-slave
        tty: true
    clang10-jnlp-slave-x86:
        build:
            context: jenkins-jnlp-slave
            dockerfile: Dockerfile
            args:
                SOURCE_CONANIO_IMAGE: conanio/clang10-x86
        image: "${DOCKER_USERNAME}/clang10-jnlp-slave-x86:${DOCKER_BUILD_TAG}"
        container_name: clang10-jnlp-slave-x86
        tty: true
```

Croydon: Does this Jenkins thing really require a 32 bit container? I was hoping we do not need this anymore for new compiler versions

uilianries: Yes, it's internal requirement for JFrog I think. But yeah, x86 it's a pain.

Croydon: Who can clarify the requirements for this? @lasote maybe?

uilianries: Maybe @danimtb or @jgsogo have that information.

jgsogo: Having a look at the first PR for these images (https://github.com/conan-io/conan-docker-tools/pull/120), probably @maheshjfrog can say something about it.

maheshjfrog: I am not familiar with requirements for individual conanio/* docker images (like the choice of 32 bit containers).
My support was to add jnlp slave jar to all existing conanio docker images so our jenkins pipeline can use them as nodes.

jgsogo: I see, then we need all these `jnlp` _duplicated_ nodes.

Croydon: Thanks for the information

Based on everything, I'm assuming that we don't need 32-bit containers for new compiler versions anymore. And if so, we could still add them later on

Croydon:
Ready for review :)

Croydon:
I really would prefer not to. As explained in the PR description, choosing a somewhat older distribution gives us an somewhat older glibc version and therefore better compatibility with many supported distribution versions out there. If we always choose the newest distribution versions for the newest compilers we will never get good binary compatibility.

uilianries:
I would prefer updating for LTS at least. Indeed glibc versions are a problem, but only for installers, which should be solved by the new cross-building feature.

Croydon:
Bionic is version 18.04 and a LTS too.
20.04 just got released today. It could not be more cutting edge than that 😄  

Which part of the new cross-building feature would solve the glibc compatibility?

uilianries:
> Which part of the new cross-building feature would solve the glibc compatibility?

build requirements no longer need arch_build and os_build, also all package will be unified, which, we will have one package per compiler version.

uilianries:
> Bionic is version 18.04 and a LTS too.

Some APT packages are important, like libc++, and they change from distro to distro.

Croydon:
> build requirements no longer need arch_build and os_build, also all package will be unified, which, we will have one package per compiler version.

But this still does not solve the problem. Somebody can use a new(er) compiler version on LTS versions like RHEL, CentOS, Ubuntu LTS etc. Their Conan instances would then try to get the builds from us which we build with newer glibc versions.

The only way I see to reduce this problem is to use older glibc versions, hence, in this case, older Ubuntu versions

uilianries:
I think again this propose and I have to agree that Ubuntu 18.04 LTS is the best choice for this case. It's a long term version and Clang 10 can be used normally

uilianries:
Great job @Croydon !

PR #194, opened 2020-04-17
Update CMake to 3.18.2

Croydon:
Changelog: Feature: Update CMake to 3.18.2

### DIFF
--- a/clang_10/Dockerfile
+++ b/clang_10/Dockerfile
@@ -10,7 +10,7 @@ ENV LLVM_VERSION=10 \
     PYENV_ROOT=/opt/pyenv \
     PYTHON_VERSION=3.8.2 \
     PATH=/opt/pyenv/shims:${PATH} \
-    CMAKE_VERSION=3.16.6
+    CMAKE_VERSION_FULL=3.18.2
 
 RUN dpkg --add-architecture i386 \
     && apt-get -qq update \
@@ -76,7 +76,7 @@ RUN dpkg --add-architecture i386 \
     && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install ${PYTHON_VERSION} \
     && pyenv global ${PYTHON_VERSION} \
     && pip install -q --upgrade --no-cache-dir pip \
-    && pip install -q --no-cache-dir conan conan-package-tools cmake==${CMAKE_VERSION} \
+    && pip install -q --no-cache-dir conan conan-package-tools cmake==${CMAKE_VERSION_FULL} \
     && chown -R conan:1001 /opt/pyenv \
     # remove all __pycache__ directories created by pyenv
     && find /opt/pyenv -iname __pycache__ -print0 | xargs -0 rm -rf \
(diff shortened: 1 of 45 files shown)


memsharded:
Please @uilianries @SSE4 I would like to make sure that the cmake version will not affect the binaries we create in ConanCenter, and that the packages in ConanCenter can be built and consumed with other older CMake versions, and not require only the very latest CMake.

Croydon:
Actual minimal CMake version: `max(recipe minimal version, project minimal version)`

Building: 
  * when it comes to the CMake files we write for the recipes we are pretty conservative. So the actual minimal CMake version will come in most cases from the projects package
  * our CMake files are rather small, simple and conservative, but it is also true that we never tested building recipes with older CMake versions. This would be pretty complex, as we would need to determinate the actual minimal CMake version and install this version on the fly in the build environment. This does not look feasible
  * at least so far we did not run in any issues which updating CMake rather quickly and I don't remember any complains
  * we probably do not need to be **THAT** quick in updating CMake, but some projects are relative quick in requiring them. So we can likely wait longer, but if we wait too long our CMake version might be too old to build newer project versions

Consuming:
  * Sometimes we need to package upstream CMake files in our packages, but then again, we do not define this minimal CMake version
  * In the other cases, Conan is supposed to generate all (or the remaining) CMake files so Conan itself is determinating the minimal version

Am I missing something?

SSE4:
I have the same feeling, we can't use old version of CMake on CCI, because libraries tend to use various new features from recent CMake releases.
that our recipes do is very simple injection, so it's only matter of that libraries are using.

Croydon:
GCC 4.6 and 4.9 fail with

> cp: cannot overwrite non-directory '/usr/share/vim/vimfiles' with directory 'cmake-3.18.2-Linux-x86_64/share/vim/vimfiles'

which I do not really understand

And "Docker Conan Tests" fail too, but I don't see any error message there. And in general I find the test logic rather hard to follow

@uilianries Could you perhaps have a look, at least for the Docker Conan Tests failure?

uilianries:
@Croydon I can't reproduce that error too. It works for me. I just restarted that job and failed again. Azure shows me a warning message about their service: https://status.dev.azure.com

I don't know if it's affecting us or not. I'll try something different, split the recipe step-by-step and see what is failing exactly.

uilianries:
@Croydon I sent a hotfix on https://github.com/Croydon/conan-docker-tools/pull/3!

uilianries:
MinGW failed, restarting ...

Croydon:
Hurrah 🥳 

Ready for review

uilianries (approved):

danimtb (approved):

uilianries:
@Croydon Kitware provides an APT repository. I think we could adopt it in the future and install the latest version available for every image, so we don't need to update from time to time, as we usually have one new release monthly.

Croydon:
The versions available on pip sometimes do lag a few versions behind, but I don't think that this has a negative impact on us. If we want to get rid of manually updating, we can remove the fixed versions in the Dockerfiles. Beyond that, I think it is more interesting to use the CMake Conan recipe and using CMake via pip for (potential) bootstrapping. Similar to what I think we will do with compilers in the future.

For now, are we waiting for more reviews or can this be merged?

uilianries:
@Croydon I forgot pip! Yes, much better than APT, as we don't need to add a new repo to our apt list.

PR #207, opened 2020-06-06
Add centos7-gcc9 docker image

### LINKED ISSUE #206: Add centos 7 with gcc 9 support
I'd like to contribute an image for the above configuration

Centos6 will no longer receive maintainance updates from November 30th, 2020. Where I work we have effectively stopped supporting it. Having an updated image would help.

rockdreamer:
Changelog: Feature: Add support for a centos7 image with GCC 9 compiler
closes #206 

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ GCC>=5 is ABI compatible for minor versions. To solve multiple minors, there are
 | [conanio/gcc9: gcc 9](https://hub.docker.com/r/conanio/gcc9/)                      | x86_64  |  Supported           |
 | [conanio/gcc9-armv7: gcc 9](https://hub.docker.com/r/conanio/gcc9-armv7/)          | armv7   |  Supported           |
 | [conanio/gcc9-armv7hf: gcc 9](https://hub.docker.com/r/conanio/gcc9-armv7hf/)      | armv7hf |  Supported           |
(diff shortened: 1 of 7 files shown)


uilianries:
Thanks for you effort! But, unfortunately it can not be accepted.

CentOS6 image was created to solve a specific problem: Provide a very old glibc version when building installer packages. When it was created, CentOS6 was totally outdated, some packages we got from the vault.

Adding a new CentOS image will increase our maintenance, and we don't have intention to keep support for RH/CentOS images, otherwise, we would have to support other distros (Fedora, Arch, ...).

Thanks again for your contribution, it seems be useful for other users which are look for a new version CentOS version.

dheater:
> Adding a new CentOS image will increase our maintenance, and we don't have intention to keep support for RH/CentOS images, otherwise, we would have to support other distros (Fedora, Arch, ...).

I think the goal here based on the referenced issue (#206) is not support for CentOS per-se, or other distros, but support for older `glibc`
Related to #204

dheater:
@uilianries I'm requesting this be reconsidered. The `centos6-gcc7` image solved a specific problem for those of us supporting distros with older `glibc` (2.17 for RHEL/CentOS7, 2.22 for SUSE 12).  Now that CentOS 6 is EOL, I think this is needed as a replacement for this use case.

PR #214, opened 2020-08-11
Bump version to 1.28.1

Croydon:
Changelog:  Feature: Bump Conan version to 1.27.1

// @uilianries 

To unblock https://github.com/conan-io/conan-center-index/pull/2037 (and probably others)

### DIFF
--- a/conan_version.txt
+++ b/conan_version.txt
@@ -116,3 +116,4 @@ Bumped for Conan 1.26.0
 
 Bumped for Conan 1.27.1
 
+Bumped for Conan 1.28.1


uilianries:
Thanks @Croydon !

ericLemanissier:
mm, don't we need to keep the conan version in sync with what is deployed on CCI's windows and mac agents ?https://github.com/conan-io/conan-center-index/wiki/Deployment-Changelog

uilianries:
Yes, but are made separated as CCI don't use the same solution for all platforms.

danimtb:
@ericLemanissier for future reference, check here https://github.com/conan-io/conan-center-index/blob/master/docs/changelog.md

ericLemanissier:
Yes, I know about this. I sent my message before https://github.com/conan-io/conan-center-index/pull/2490 was created. The point I'm trying to make is that the person in charge of upgrading the conan version deploy on CCI's agents should probably bump the version in conan docker tools images at the same time, or it will certainly be forgotten at some point. Or maybe it is already the case, and Croydon just happened to be faster than this person of few hours ?

danimtb:
currently, it is a manual task, but we will try to sync as much as possible and we are working on decoupling conan-docker-tools updates  from the images used in CCI. Thanks a lot for pointing out 👍

Croydon:
> we are working on decoupling conan-docker-tools updates from the images used in CCI

You mean that the images used in CCI in the future won't be the same as here?

danimtb:
no, I am saying that the update of the Conan version in this repo won't be coupled to updating it in CCI. Basically CCI is using latest images instead of a pinned version

PR #219, opened 2020-09-10
Bump version to 1.29.1

ericLemanissier:
Changelog: Feature: Bump Conan version to 1.29.1

this will unblock at least the following:
- https://github.com/conan-io/conan-center-index/pull/2741
- https://github.com/conan-io/conan-center-index/pull/2303
- https://github.com/conan-io/conan-center-index/pull/2301

### DIFF
--- a/conan_version.txt
+++ b/conan_version.txt
@@ -117,3 +117,5 @@ Bumped for Conan 1.26.0
 Bumped for Conan 1.27.1
 
 Bumped for Conan 1.28.1
+
+Bumped for Conan 1.29.1


ericLemanissier:
1 job errored with `ERROR: Error 500 downloading file https://github.com/google/googletest/archive/release-1.8.1.tar.gz`
Probably you can use "Rerun failed jobs" button ?

uilianries:
Finally they added re-run failed jobs!

@ericLemanissier thanks for your PR. We have to wait more 1 or 2 weeks before merging it. As CCI consumes the latest Docker image available, it will create a problem if merged now. In the future it should be fixed for CCI.

Croydon:
> As CCI consumes the latest Docker image available, it will create a problem if merged now.

What problem would this cause?

uilianries:
We can't different versions at same time. Windows and OSX builders are machines, so we need to update them all same time. Imagine a recipe which requires Conan 1.29 failing on Windows because Conan is outdated, but it pass on Linux.

uilianries (approved):

PR #221, opened 2020-09-21
Add Docker tests agent image

uilianries:
Changelog: Feature: Add Conan Test + Jenkins client for internal usage.

closes #220

- [ ] Refer to the issue that supports this Pull Request.
- [ ] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [ ] I've read the [Contributing guide](https://github.com/conan-io/conan-docker-tools/blob/master/.github/CONTRIBUTING.md).
- [ ] I've followed the [PEP8](https://www.python.org/dev/peps/pep-0008) style guides for Python code.
- [ ] I've followed the [Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) guides for Dockerfile.

PR #223, opened 2020-09-23
Add Docker for jenkins agent

### LINKED ISSUE #222: Add docker to conantestagent image
For this image we need to be able to run docker and docker-compose to use it in Conan CI so it should have docker installed.
Besides the infrastructure part, this should be a matter of just doing these steps when creating the image? https://docs.docker.com/engine/install/ubuntu/

uilianries:
closes #222

Changelog: Feature: Docker Jenkins Agent is able to run Docker

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,6 +15,10 @@ jobs:
       DOCKER_PASSWORD: $(DOCKER_PASSWORD)
   strategy:
     matrix:
+      Conan Tests Agent:
+        BUILD_CONAN_TEST_DOCKER: 1
+        BUILD_CONAN_TEST_AGENT: 1
+
       CentOS 6 GCC 7 x86_64:
         GCC_VERSIONS: "7"
         DOCKER_ARCHS: "x86_64"
@@ -167,7 +171,3 @@ jobs:
       Conan Tests:
         BUILD_CONAN_TESTS: 1
         BUILD_CONAN_TEST_AZURE: 1
-
-      Conan Tests Agent:
-        BUILD_CONAN_TEST_DOCKER: 1
-        BUILD_CONAN_TEST_AGENT: 1
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -169,5 +169,5 @@ jobs:
         BUILD_CONAN_TEST_AZURE: 1
 
       Conan Tests Agent:
-        BUILD_CONAN_TESTS: 1
+        BUILD_CONAN_TEST_DOCKER: 1
         BUILD_CONAN_TEST_AGENT: 1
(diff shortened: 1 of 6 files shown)


azure-pipelines.yml
```diff
@@ -15,6 +15,10 @@ jobs:
       DOCKER_PASSWORD: $(DOCKER_PASSWORD)
   strategy:
     matrix:
+      Conan Tests Agent:
+        BUILD_CONAN_TEST_DOCKER: 1
```
uilianries, line 19: order is not important, only to check fist in PR result, otherwise will take 1h

conan_test_docker/Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM conanio/gcc5
```
uilianries, line 1: copy of conan_tests + Docker package installation and system docker enable.

More information  how to install on linux: https://docs.docker.com/engine/install/ubuntu/

czoido (approved):

danimtb (changes requested):
I don't understand why we are creating a new image with this changes. I thought that we were going to add all the extra needed things (Jenkins slave + docker, at the moment) to just a new image called "conantestagent". Why do we need an extra "conantestdocker"?

uilianries:
@danimtb As @czoido asked, a new Docker image is required with Jenkins client + Docker inside.

For now we have 2 basic images, conan_tests which only contains gcc5 and some python versions installed, and Jenkins client image, which contains only stuff to run Jenkins jobs.

To run Docker, we need to install of course, so we have few options, or add Docker by default in conan_tests, or create a new image as extra layer. This PR follows the second option. We could use the first, but it will increased conan_tests image size.

danimtb:
The only thing we should care is not braking our current test environment, to it is oke to have the old image untouched and have a new one with the new requirements. In this case the new requirements are jenkins + docker. I think there is no reason to have more images than those 2 IMO

uilianries:
THe last PR https://github.com/conan-io/conan-docker-tools/pull/221/files there was no new image, just a connection between conan_tests and jenkins image. As Jenkins requires a base image, we use conan_tests. So conantestagent is just a merge, not a new recipe indeed.

Now it requires a new package, Docker, docker-compose and service upstart. I could rename conantestdocker to conantestagent and install docker packages as well. Otherwise, we will need to install in conantests or in jenkins client.

uilianries:
@czoido @danimtb Now there is the real test_agent image, which is the combination of conantest + jenkins + docker. It's under jenkins folder because it requires some copies. The old conantestdocker has been deleted.

czoido:
Looks great! Thanks a lot!

czoido (approved):

danimtb (approved):
Much better! thank you for the clarification and the quick changes

PR #234, opened 2020-12-10
fix gcc and clang 9 images

ericLemanissier:
eoan is EOL
fixes #233

Changelog: (Feature | Fix | Bugfix): Describe here your pull request

- [X] Refer to the issue that supports this Pull Request.
- [ ] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [X] I've read the [Contributing guide](https://github.com/conan-io/conan-docker-tools/blob/master/.github/CONTRIBUTING.md).
- [ ] I've followed the [PEP8](https://www.python.org/dev/peps/pep-0008) style guides for Python code.
- [ ] I've followed the [Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) guides for Dockerfile.

PR #235, opened 2020-12-10
Replace eoan by Bionic

uilianries:
Ubuntu Eoan (19.10) reached EOL and archives.ubuntu.com won't work. Another option is adding old-releases.ubuntu.com, but it didn't work for me.

Ubuntu Bionic (18.04) is LTS and will be supported until 2024. As it's older than 19.04, the glibc version is compatible. 

Ubuntu only provides a single package version, so if a package becomes outdated, won't be possible to install the old version, thus specific version or ranged version is useless.

fixes https://github.com/conan-io/conan-docker-tools/issues/233 

### DIFF
--- a/clang_9-x86/Dockerfile
+++ b/clang_9-x86/Dockerfile
@@ -1,4 +1,4 @@
-FROM i386/ubuntu:eoan
+FROM i386/ubuntu:bionic
 ENTRYPOINT ["linux32", "--"]  # https://github.com/conan-io/conan-docker-tools/issues/36
 
 LABEL maintainer="Luis Martinez de Bartolome <luism@jfrog.com>"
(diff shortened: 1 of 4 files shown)


ericLemanissier:
I think we have no choice but keeping the exact same version of glibc, otherwise there will be problems when running a test recipe in this container, with required libraries downloaded from CCI which were compiled previously on eoan, and use newer glibc than bionic

uilianries:
Bionic is older than eoan and glibc is backward compatible.

uilianries:
Well, Bionic doesn't offer gcc9

ericLemanissier:
all gcc 9 binaries currently on CCI use glibc 2.30. Bionic has glibc 2.27, so it won't be able to load these existing binaries. Or do you suggest that C3I has the possibility to regenerate all gcc 9 binaries of all recipes, so that these binaries can run on bionic ?
Anyway, this is irrelevant now that we know bionic does not offer gcc9. So, we are doomed ?

jgsogo:
`old-releases.ubuntu.com` have to work... maybe there is a timing issue until they are available worldwide (I detected the failure this afternoon)? 🤷  **We really need them to work**

---

> Or do you suggest that C3I has the possibility to regenerate all gcc 9 binaries of all recipes, so that these binaries can run on bionic ?

@ericLemanissier not yet, we would love to (and we will have to) but it is not going to be easy.

Croydon:
Can be closed in favour of https://github.com/conan-io/conan-docker-tools/pull/234

uilianries:
Fixed by #234

PR #236, opened 2020-12-11
Deprecate CentOS 6

uilianries:
Another distro reached EOL and I didn't find an `old-releases` equivalent for CentOS6. This situation is bad, because now we can't update our images based on CentOS, making other PRs broken.

My idea is deprecate Cent OS 6 on this PR, ~and open a new one replacing by Cent OS 7.~

Cent OS 6 uses glibc 2.12
Cent OS 7 uses glibc 2.17

/cc @ericLemanissier 

UPDATE:

* Add CentOS 7 is no longer needed, as CCI (official) doesn't use those images, only based on Ubuntu.


Croydon (approved):
I'm not sure if we actually want to add another CentOS based image. I think in the long-term it would be awesome if we could build our compilers ourselves (hello GCC and LLVM Conan packages!) and then choose an actively supported LTS distro as a base image for all containers.

The huge advantage for CentOS was the old glibc versions, making binaries compatible with pretty much every supported distribution, but we don't use it in CCI anyway and I [just migrated Bincrafters away from it too](https://github.com/bincrafters/bincrafters-package-tools/commit/ba165ad14c7090f7d47c2500ae1fd97a834935de)

jgsogo (approved):

uilianries:
> I'm not sure if we actually want to add another CentOS based image. I think in the long-term it would be awesome if we could build our compilers ourselves (hello GCC and LLVM Conan packages!) and then choose an actively supported LTS distro as a base image for all containers.
> 
> The huge advantage for CentOS was the old glibc versions, making binaries compatible with pretty much every supported distribution, but we don't use it in CCI anyway and I [just migrated Bincrafters away from it too](https://github.com/bincrafters/bincrafters-package-tools/commit/ba165ad14c7090f7d47c2500ae1fd97a834935de)

Same feeling. We don't use for CCI too, the only propose would be for community, but we don't want to maintain it we don't use directly. Indeed, adding compiler packages is the future IHMO.

Issue #240, opened 2021-01-04
Are Clang 3.8, GCC 4.6 and and GCC 4.8 images still used?


Croydon:
Are Clang 3.8, GCC 4.6 and and GCC 4.8 images still used? Do they have known users?

My question is basically if it is worth to continue maintaining them

PR #241, opened 2021-01-05
Eating Our Own Ice Cream: Use Conan's CMake recipe

Croydon:
Changelog: Feature: Use a Conan build of CMake for our build images

https://en.wikipedia.org/wiki/Eating_your_own_dog_food#Alternative_terms 

Let's try this first with CMake and get some experience. Later we can 
  * move more dependencies like this.
  * Particular the compilers itself once GCC and LLVM recipes are ready for that. This would makes us vastly more independent of specific Ubuntu versions.
  * @madebr is working on a Python recipe, maybe this will work well too ✌️ 

_____________________

- [x] Requires https://github.com/conan-io/conan-center-index/pull/4147

_____________________

Some explanations:
  * GCC 4.6 is too old to build CMake. Hence we download the official binaries von Kitware. This works because this image is 64 bit and there is no GCC 4.6 32 bit container.  Kitware does not provide 32 bit binaries. (And these old containers should probably be removed anyways, https://github.com/conan-io/conan-docker-tools/issues/240)
  * Our CMake recipe is building CMake with CMake
  * Bootstraping of CMake works via `make`
  * For most other compiler versions, except the newest, we do have 32 bit containers. Here, we can not use the official binaries, but rather we compile the source code and bootstrap a CMake build via `make`. 
  * Under regular circumstances, we already have Conan packages for CMake available which can be just downloaded. However, who knows what could happen, so these containers should be able to bootstrap. Only when the Conan remote offers no CMake packages to download, we build the Conan CMake recipe with the CMake binary we acquired previously
  * The final image only contains a CMake which was build via Conan (except the GCC 4.6 image)
  * In the future, we might want to add a bootstrap option directly to the CMake Conan recipe, but for now, this is fine
  * This also works good as a proof-of-concept and this workflow can be used eventually for the compilers, the Python runtime and probably other packages too. But let's see how this performs in practice with CMake as this is ready-to-go right now and much less complex than those other packages

### DIFF
--- a/clang_9/Dockerfile
+++ b/clang_9/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:eoan
+FROM ubuntu:eoan AS basics
 
 LABEL maintainer="Luis Martinez de Bartolome <luism@jfrog.com>"
 
@@ -56,15 +56,6 @@ RUN dpkg --add-architecture i386 \
     && update-alternatives --install /usr/bin/cpp cpp /usr/bin/clang++-9 100 \
     && ln -s /usr/include/locale.h /usr/include/xlocale.h \
     && rm -rf /var/lib/apt/lists/* \
-    && wget -q --no-check-certificate https://cmake.org/files/v${CMAKE_VERSION_MAJOR_MINOR}/cmake-${CMAKE_VERSION_FULL}-Linux-x86_64.tar.gz \
-    && tar -xzf cmake-${CMAKE_VERSION_FULL}-Linux-x86_64.tar.gz \
-       --exclude=bin/cmake-gui \
-       --exclude=doc/cmake \
-       --exclude=share/cmake-${CMAKE_VERSION_MAJOR_MINOR}/Help \
-       --exclude=share/vim \
-    && cp -fR cmake-${CMAKE_VERSION_FULL}-Linux-x86_64/* /usr \
-    && rm -rf cmake-${CMAKE_VERSION_FULL}-Linux-x86_64 \
-    && rm cmake-${CMAKE_VERSION_FULL}-Linux-x86_64.tar.gz \
     && wget -q --no-check-certificate -O /usr/local/bin/jfrog https://api.bintray.com/content/jfrog/jfrog-cli-go/\$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64 \
     && chmod +x /usr/local/bin/jfrog \
     && groupadd 1001 -g 1001 \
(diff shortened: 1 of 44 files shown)


Croydon:
Ready!

clang_9/Dockerfile
```diff
@@ -90,9 +81,35 @@ RUN dpkg --add-architecture i386 \
...
+    && make -s -j`nproc` \
+    && make -s install > /dev/null \
+    && cd - \
+    && rm -rf cmake-* \
+    # Install Packages which we want to deploy
+    # Install CMake + move license files to specific licenses/cmake subdirectory for deploy
+    && conan install cmake/${CMAKE_VERSION_FULL}@ --build missing -g deploy \
+    && cd cmake/licenses \
+    && mkdir cmake \ 
+    && ls | grep -v cmake | xargs mv -t cmake
```
madebr, line 103: I don't completely understand everything you do here, but would it also be possible to
use a `conanfile.py` script with a `deploy` method here?
When/if multiple conan packages will get pre-installed, then it would be a matter of updating the conanfile, instead of all dockerfiles.

Croydon: The CMake version is already in every `Dockerfile`. 

And in the feature we will have much less `Dockerfile`s to maintain either due to something like #204 or by using `ARGS` for values which differ between the `Dockerfile`s, or both. The second thing is something I would like to check out next, as mass editing of so many  `Dockerfile`s continuously lead to errors.

Either way, currently there are differences. For instance, in the GCC 4.8 image we always have to rebuild CMake, as the package is build in the GCC 4.9 image with a newer GLIBC/GLIBCXX versions and can not run on the GCC 4.8 image.

Newer CMake versions might also not compile anymore in old compiler. And thinking ahead, I believe newer Python versions already have issues with too old distro versions IIRC. 🤔  

All in all, I don't think a single `conanfile.py` would really simplify much right now, but might be worth considering in the future again

Croydon:
May I get some feedback here? @uilianries @danimtb @czoido 😃

uilianries (approved):
LGTM I like this idea

jgsogo (changes requested):
Hi! I'm a bit concerned about these changes to these images.

I totally agree we should eat our own dog food and, if these docker images were distributed ONLY to outside parties, I'd totally support installing these tools using Conan. It totally makes sense: Conan generates docker images and uses Conan to install tools 💯 

BUT, these images are used in `conan-center-index` too, these images are the ones we use to compile and generate packages, this CMake is the one we use to compile the packages (`cmake/3.18.5` too). If we break `cmake` package and update the docker image, _we can't recover_, we won't be able to compile CMake in `conan-center-index` because the installed CMake is broken.

This is very unlikely, but there is a risk (and some headache) we can avoid. We can talk about this, but we need to be sure about what we are doing and why, it requires some consensus. Maybe we need to differentiate between the images we use in C3i (use an independent mechanism to install tools) and other ones we could start to _distribute_ as "Conan images ready to build your binaries".

uilianries:
I like how @jgsogo predict possible failures involving CCI and CDT, indeed the risk is real.

What about installing a specific package revision? So we could prevent any possible broken package.

uilianries:
@Croydon Sorry for creating a conflict with your PR :disappointed:

Croydon:
I think splitting containers would be the worst possible thing we could do, so personally, I'm against that.

You have a good point to worry about a dead lock situation.

Some ideas:
  * we leave it as it is, if
    *  a new recipe revision breaks, we can update the containers by specifying a specific revision
    * the entire Conan remote breaks, we could still check out the recipe from git and leave the rest unchanged
  * we specify right now a specific recipe revision, however, that would mean constant updating work and might be super annoying and confusing. To minimize this, we might want to only update the revision when updating the CMake version itself
  * we check out the recipe right now from git, but this might be an unnecessary complication for an unlikely dead lock situation, in which case we could still do this

Croydon:
I pinned the CMake recipe revision

Croydon:
As another note, it is pretty unlikely that we deploy new Docker images with a broken CMake since the CI here is building Conan packages with CMake in the tests

Croydon:
Closing this as it would require a complete rewrite by now.

However, some of the concepts are merged by now, like using different container layers, so that temporary files don't get into the final container, using more parameters to configure things and starting to use Conan packages for creating the containers itself (https://github.com/conan-io/conan-docker-tools/pull/370) 👏

uilianries:
@Croydon Thank you for sharing this idea. Indeed we started to think about using Conan packages instead of APT packages on "modern" Docker images. Binutils is a good start because we didn't find the latest version available from an external PPA and building from source would consume even more time when creating a Docker image. As next step, probably we will internalize other tools which are available in Conan Center, or convert them to Conan packages, e.g. git.

PR #243, opened 2021-01-19
Install JFrog CLI for all CCI images

uilianries:
As Conan Center Index is doing internal changes, JFrog CLI will be required.

I didn't add for x86 images because they are optional and not used by CCI.

### DIFF
--- a/build.py
+++ b/build.py
@@ -291,6 +291,14 @@ class ConanDockerTools(object):
                 "arch_build=%s -s os_build=Linux --build" % (self.service, arch),
                 shell=True)
 
+        try:
+            subprocess.check_call(
+                "docker exec %s ls /usr/local/bin/jfrog" % self.service, shell=True)
+        except:
+            pass
+        else:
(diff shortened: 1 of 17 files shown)


jgsogo (changes requested):
We should better use documented install command: https://jfrog.com/getcli/ 

```
wget -qO - https://releases.jfrog.io/artifactory/api/gpg/key/public | sudo apt-key add -;
echo "deb https://releases.jfrog.io/artifactory/jfrog-debs xenial contrib" | sudo tee -a /etc/apt/sources.list;
sudo apt update;
sudo apt install -y jfrog-cli;
```

Croydon:
It only provides debs for bionic, eon, focal and xenial, so it won't work for all images

uilianries:
Yes, that's why I downloaded directly. 

The script provided by Jfrog runs wget, but downloads to the current folder.
Basically, I'm running GetCi.sh from JFrog

jgsogo (commented):
👍

danimtb (approved):

PR #245, opened 2021-02-09
Bump Conan version to 1.33.1

danimtb:
Changelog: Feature: Bump Conan version to 1.33.1

### DIFF
--- a/conan_tests/Dockerfile
+++ b/conan_tests/Dockerfile
@@ -8,11 +8,13 @@ RUN sudo apt-get update \
     && sudo apt-get update \
     && sudo apt-get -qq install -y --no-install-recommends \
     python-software-properties \
+    python3.5 \
     python3.6 \
     python3.7 \
     python3.8 \
     python-setuptools \
     python-dev \
+    python3.5-dev \
     python3.6-dev \
     python3.7-dev \
     python3.8-dev \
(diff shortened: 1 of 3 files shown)


uilianries (approved):

conan_tests/Dockerfile
```diff
@@ -24,12 +22,14 @@ RUN sudo apt-get update \
...
     && wget --no-check-certificate --quiet -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py \
-    && sudo python3 /tmp/get-pip.py \
-    && sudo pip install "virtualenv<20.0.0" \
-    && pip3 install PyGithub \
-    && pip3 install meson
+    && sudo python3.8 /tmp/get-pip.py \
+    && sudo python3.8 -m pip install --upgrade pip \
+    && sudo python3.8 -m pip install "virtualenv<20.0.0" \
+    && sudo python3.8 -m pip install PyGithub \
+    && sudo python3.8 -m pip install meson
```
Croydon, line 32: Why installing meson? We are adding meson as a build dependency to all recipes

uilianries: I don't know, it was already there when I arrived. :stuck_out_tongue_closed_eyes: But indeed it doesn't make sense as we have Menson installer.

uilianries: I'll remove it on next commit. Azure test image is broken too. get-pip now uses f-string, which means /usr/bin/python3 must be 3.6 or higher

uilianries: @Croydon now I see: https://github.com/conan-io/conan-docker-tools/pull/175 introduces Meson, but it's for internal tests or to run another stuff internally in CCI. I won't remove now because probably it will affect CCI directly. :sweat:

PR #246, opened 2021-02-09
Fix Conan Test bad imports

uilianries:
Unfortunately, #245 broke ConanCI tests.

We needed to update conanio/conantests because get-pip.py now uses f-string (since 3.6), but the default version was 3.5.

Also, conantests has a mix of python versions from APT and pyenv. This mess caused a confusion in my mind. I forced /usr/bin/python3.8 as default version, but actually the default python runner is provided by pyenv (3.5.7).

As we are running pyenv, I believe those python versions from APT are unnecessary, we can install all them from pyenv and activate when necessary. 

From now, it will run Python 3.8.1 by default. But older versions are all installed too.

/cc @Croydon 

### DIFF
--- a/conan_tests/Dockerfile
+++ b/conan_tests/Dockerfile
@@ -9,8 +9,13 @@ RUN sudo apt-get update \
     && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.5.7 \
     && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.6.9 \
     && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.8.1 \
-    && pyenv global 3.8.1 \
-    && python -m pip install --upgrade pip "virtualenv<20.0.0" PyGithub conan meson
+    && pyenv global 3.8.1
+
+RUN python -m pip install --upgrade pip
+RUN python -m pip install meson
+RUN python -m pip install "virtualenv<20.0.0"
+RUN python -m pip install PyGithub
+RUN python -m pip install conan
 
 USER conan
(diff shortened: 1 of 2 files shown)


conan_tests/Dockerfile
```diff
@@ -3,33 +3,14 @@ FROM conanio/gcc5
...
-    && sudo ln -sf /usr/bin/python3.8 /usr/bin/python3 \
-    && wget --no-check-certificate --quiet -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py \
-    && sudo python3.8 /tmp/get-pip.py \
-    && sudo python3.8 -m pip install --upgrade pip \
-    && sudo python3.8 -m pip install "virtualenv<20.0.0" \
-    && sudo python3.8 -m pip install PyGithub \
-    && sudo python3.8 -m pip install meson
+    && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 2.7.16 \
+    && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.5.7 \
+    && PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.6.9 \
```
uilianries, line 10: Don't need 3.7.5, already installed from conanio/gcc5

SSE4 (changes requested):
may you add some tests to prevent such breakage in future?

czoido:
Thanks a lot @uilianries, I have ported these changes to https://github.com/conan-io/conan_ci_jenkins/tree/master/docker_images/conantests now we can remove the image from this repo or rename it. It will be uploaded to dockerhub from conan_ci_jenkins

PR #248, opened 2021-02-17
Add pkgconfig for clang images

uilianries:
Add pkg-config for Clang 10 and 11.

### DIFF
--- a/clang_10/Dockerfile
+++ b/clang_10/Dockerfile
@@ -24,6 +24,7 @@ RUN dpkg --add-architecture i386 \
        gnupg \
        ca-certificates \
        dh-autoreconf \
+       pkg-config \
     # Install compiler toolset
     && apt-get install -y --no-install-recommends \
        lsb-release \
--- a/clang_11/Dockerfile
+++ b/clang_11/Dockerfile
@@ -25,6 +25,7 @@ RUN dpkg --add-architecture i386 \
        gnupg \
        ca-certificates \
        dh-autoreconf \
+       pkg-config \
     # Install compiler toolset
     && apt-get install -y --no-install-recommends \
        lsb-release \


danimtb (approved):

czoido (approved):

jgsogo (approved):

Croydon:
Why? We are adding pkgconf as a build_require to recipes

Croydon:
It was on purpose that it was not added to newer images, similar like ninja is not added to newer images. Removing from old images might break users, but with new images we don't have to worry about this.

jgsogo:
Hi, @Croydon. We will start to use these images in CCI and we need them to be _equal_ to the ones we are already using for the rest of the compilers... otherwise we will get some _unexpected_ errors when submitting PRs that were not failing before and now they will start to fail just because we are using some new profiles (which is something _transparent_ to the users).

We really need all the docker images to be consistent. This is important. Due to this issue, now I don't know if some of the references that have failed when building the new configurations have failed because of this `pkg-config` issue (like `glew`) or have failed because they are not working for the new compilers. Now I need to trigger the build of all those references again because some of them will succeed now that `pkg-config` is available and, if they succeed, all their consumers now have a second chance to succeed...

---

Said that, **if we want to remove `pkg-config` from the images we can consider it**, but first we need to do an effort to add it as a `build_requires` to all (most) of the recipes. One of the things we need to tackle next is the `glibc` issue, it will require us to change all the docker images and we will need to regenerate lots of binaries. In this process maybe we can remove pkg-config from the system (all images or none).

I still don't know how to change those docker-images, but we will need to provide a way to change them in order (graph dependency order) so we need to have two sets of images: the new ones using LTS Ubuntu version (without pkg-config) and the old ones that are using _latest_ Ubuntu. Step by step, library by library, we need to use the new images, and we need to tell the CI that PRs to some packages should use new images, while other PRs should use the old images. Probably we will need to maintain the list of packages in a file.

Of course, this slow process won't prevent us from adding `pkg-config` as build-requires to recipes that use it. This slow process will only force it, and that's the reason why we need to do it carefully. 

--- 

Hope it makes sense. It doesn't mean we want pkg-config installed at system level, it just mean that we need all the images to be equal.

Croydon:
Thanks for the explanation, that makes sense.

PR #249, opened 2021-02-18
Validate essential packages

uilianries:
The PR #248 was fix because we didn't check essential packages present in all Docker images. Some packages system required, like wget and curl, but others a CCI requirement, like jfrog-cli.

/cc @Croydon 

### DIFF
--- a/build.py
+++ b/build.py
@@ -182,6 +182,7 @@ class ConanDockerTools(object):
                 self.test_jenkins()
             else:
                 self.test_linux(arch, compiler_name, compiler_version, distro)
+                self.test_essential_packages()
         finally:
             subprocess.call("docker stop %s" % self.service, shell=True)
             subprocess.call("docker rm %s" % self.service, shell=True)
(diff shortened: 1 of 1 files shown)


jgsogo (approved):
👍 

And now we can start to think about how to remove `pkg-config` in such a way that we don't break existing CCI packages.

Basically two approaches:
 * we validate existing packages doesn't require `pkg-config` and just remove it.
 * we plan a smooth transition taking into account CCI graph, starting from the top of the graph to the bottom. (This kind of changes can take weeks to propagate for all the graph.)

uilianries:
I can take a look on CCI package which require pkgconf, create a list as an issue, update one-by-one and finally, remove pkg-config from docker images.

SSE4:
I think `*/system` packages need `pkg-config`, at very least

uilianries:
> I think `*/system` packages need `pkg-config`, at very least

We could install `pkgconf` package from Conan Center. However, first we need to install pkg-config, then build all minimal stuff required by base, then remove `pkg-config`, and install `pkgconf` Conan package.

Issue #205, opened 2020-06-04
Use base docker image for multiple compiler versions


uilianries:
### Description of Problem, Request, or Question

We are running around our tail when trying to release new entire docker image for each new compiler version. This situation is worst when thinking about the multiple glibc versions available and the age of some images (based on its distro and support).

This issue asks for a new approach: A base image which can be reused by any compiler version.

What's the cost? Build most of compiler from sources and increase the CI.

Pros:
- Better maintenance, fix only one image
- Easier to add a new compiler version, only update docker-compose file
- Same glibc version

What about the old images? Keep then in the repository, but do not keep the support. The new images should have a new name, to avoid any conflict.

closes #204

Croydon:
The problem is what should the base image be?

If we choose e.g. Xenial that the glib is probably newer than that of the currently RHEL supported versions (haven't checked it yet)

But if we get this sorted out that an unified base image would be wonderful obviously

uilianries:
We will rebuild the packages in CCI to avoid any problem, also, not all images will be update, it will be from GCC 6.

ericLemanissier:
so this won't solve https://github.com/conan-io/conan-center-index/issues/213, because all installer images will still be based on xenial, and gcc 4.9 will still be based on trusty. Or am I missing something ?

uilianries:
that's another case, we need to filter installer packages in CCI and build then using a specific image (old).

Croydon:
The problem seems to be that packages are first build with GCC 5, meaning that installer packages are ONLY build on GCC 5.

The glibc verison of our GCC 5 container is newer as the one in the GCC 4.9 container, so it will fail to use this executables in GCC 4.9 CI builds.

If this changes only GCC 6 and newer, then it won't solve the problem within our CI builds as @ericLemanissier pointed out.

But such problems don't happen only in our CI builds, it is the same for consumers with Linux distribution versions which use older glibc versions. 
That's why INCREASING the glibc version might actually be the opposite of what we want.

________________________________

I have no clue how hard it is to do, but just to throw this idea from my head into the wild: When we already considering building the compilers from source, maybe building a light distribution ourselves with Yocto might be a solution? In (absolute) theory, this could allows us to combine old glic versions with all compiler versions we want?

uilianries:
Again, this issue is not related to https://github.com/conan-io/conan-center-index/issues/213. Here we want to avoid new glibc version and distro versions.

[...]

PR #252, opened 2021-03-01
Install libc++-11 with clang-11

### LINKED ISSUE #251: Wrong libc++ version installed with clang-11
https://github.com/conan-io/conan-docker-tools/blob/9a26bad2285a4b87d0c20e6b664d64b7dbb1a95d/clang_11/Dockerfile#L39-L40

These two lines must be:
```
       libc++-11-dev \
       libc++abi-11-dev \
```
Otherwise, the installed version is to old (6 or 7, I don't remember). My proposal is consistent with llvm official [page](https://apt.llvm.org/#install_stable).

Related to https://github.com/conan-io/conan-center-index/pull/4630

Woazim:
Changelog: Fix: This is to install libc++-11 instead of default one (6 or 7, very old)

fixes #251

### DIFF
--- a/clang_11/Dockerfile
+++ b/clang_11/Dockerfile
@@ -36,8 +36,8 @@ RUN dpkg --add-architecture i386 \
     && rm llvm.sh \
     # Further LLVM packages not installed by the llvm.sh script
     && apt-get install -y --no-install-recommends \
-       libc++-dev \
-       libc++abi-dev \
+       libc++-11-dev \
+       libc++abi-11-dev \
     && apt-get remove -y lsb-release software-properties-common \
     # Install dependencies of Python
     && apt-get install -y --no-install-recommends \


uilianries (changes requested):
Thank you for your contribution!

Unfortunately we can't accept this change. Updating libstdc++ or libstdc++11 will break official packages in Conan Center. If you really need a newer version, we can do it locally.

We will provide new Docker images soon, including newer libc version. Thus, your hotfix is still a valid idea, but it will be included for other Docker images.

uilianries:
Looking https://github.com/conan-io/conan-center-index/pull/4630, it seems like no clang-11 package has been generated yet. Thus, it's possible to accept your PR.

uilianries (approved):
LGTM, it fixes https://github.com/conan-io/conan-center-index/pull/4630

Croydon (approved):

jgsogo (changes requested):
Need to understand the implications. If we upgrade this library to version `libc++11`, then the binaries compiled using this new version can run with the old one? Does every consumer need to upgrade to `libc++11` in order to run the binaries compiled with it?

That would be a blocker, as `libc++` will be the default for many users.

Woazim:
I think there is a misunderstanding. clang-11 should come with libcxx-11. But this lib and its ABI must be installed separately as it's noticed in official clang page [here](https://apt.llvm.org/#install_stable) (see 'To install all key packages' / '# Clang and co'. If you just install libcxx through `apt-get install libcxx-dev` you install a very old version of libcxx (version 6). This version does not support modern C++ and is not what clang-11 users expect.

By the way, this [official LLVM link](https://libcxx.llvm.org/docs/DesignDocs/ABIVersioning.html) explains that "Libc++ aims to preserve stable ABI to avoid subtle bugs when code built to the old ABI is linked with the code build to the new ABI". So it should not be a problem to correct this, but, I agree with you, it can not be sure.

Finally, this Dockerfile is not consistent with the one for clang-10 (which is good). You'll see on [these lines](https://github.com/conan-io/conan-docker-tools/blob/058db05060cfedc3f78e42320a0980e2df56d4cf/clang_10/Dockerfile#L37-L39) that libcxx-**10**-dev is installed (not libcxx-dev).

I'm sorry, but there is clearly a problem with this clang-11 installation. I don't know the right way to correct this, but it must be corrected. From your side, is it possible to rebuild all clang-11/libcxx in conan-center recipes to avoid side effects?

jgsogo:
We will be using the same docker image for all the compilers (we are proposing Ubuntu 16.04) to ensure all the binaries have the same and a lower enough `glibc` version (this is convenient for the `build_requirements`). All current images will be deprecated in favor of those ones.

The issue would be, for example, if we build some tool using clang-11 (with `libc++11`) and then we need to run that tool on the images we use for all the `gcc` and `clang` compilers... will it run?

uilianries:
I didn't find any clear information about libc++ backward compatibility, but Marshall Clow (libc++ owner) [answered](https://stackoverflow.com/questions/66477097/is-libc-backward-compatible-with-older-versions) that it should work.

Woazim:
I've run a little test: I build ImaGL and its test program with clang-11, libc++-11. I ran it on a fresh Ubuntu 16.04 with default libc++ (v3.7) without any problem.

jgsogo:
There are several questions that came to me related to this issue and they will help us a lot with other ongoing efforts as well.

 * The plan is to move to the same base docker image for all the Linux compilers, ATM the chosen one is `Ubuntu 16.04` (LTS).  We still don't know if this scenario is possible, there are some questions that need answers:
    * Can we get to run all the compilers (from GCC 4.9 to GCC 10, and clang 3.9 to 11) in this version of Ubuntu? We will need to compile from sources some of them... (versions TBD https://github.com/conan-io/conan-center-index/issues/4761)
    * We would need to compile some missing packages too: is `libc++11` available for `16.04`? can we build it from sources?

    Maybe we realize we cannot use the same base image for all the compilers and it will totally change the perspective from where to look at this issue.

 * We need to understand why the `libc++` package now is called `libc++11`, surely there is some ABI break. We just need to confirm it, it is ok to write a demo app to demonstrate it. **It won't be a blocker**, but we need to know the implications of the decision. 
   * Can `LIBCXX_ABI_VERSION` help us? Can we use version 11 but use `LIBCXX_ABI_VERSION` to target old ABIs and still get all the modern C++ features?

 * If ABI is incompatible, should we consider it as a new entry to `compiler.libcxx` for Clang compilers and add `libc++11` to the list?

uilianries:
I would to continue this discussion about unifying or not at #205, which is the appropriated issue. This PR is related to clang fix.

jgsogo:
I think I did some fundamental failure above. The runtime packages are `libc++abi1-11` and `libc++1-11` which (if I'm reading the names properly) are compatible for all versions from `1` to `11` (Ubuntu 20.04). So, they must be able to run any binary compiled up to `libc++-11-dev`.

So, this is not about installing `libc++-11` as the runtime, as a different package from `libc++`, this is only about upgrading the `dev` library... and we will need to be sure that we can install `libc++abi1-11` in the base image we want to use.

The question, for this issue, is just: are there `libc++abi1-11`/`libc++1-11` packages available for Ubuntu 16.04? If not, can we build them from sources?

Sorry for the misunderstanding.

Woazim:
> The question, for this issue, is just: **are there `libc++abi1-11`/`libc++1-11` packages available for Ubuntu 16.04? If not, can we build them from sources?**

Short answer: yes

Long answer: yes, following these instructions from a docker ubuntu:xenial image:
``` sh
apt-get update
apt-get install -y wget software-properties-common apt-transport-https
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
add-apt-repository "deb http://apt.llvm.org/xenial/   llvm-toolchain-xenial-11  main"
apt-get update
apt-get install -y libc++1-11 libc++abi1-11
```

By the way, libc++1-12 and libc++abi1-12 from clang-12 are also installable with this script.

jgsogo:
Please, have a look to this approach: https://github.com/conan-io/conan-docker-tools/pull/267

ericLemanissier:
This can now be closed

PR #265, opened 2021-06-01
[hotfix] Build GCC11 in CI

uilianries:
GCC 11 integrated to the repository, but not listed to be built in CI.

This PR also updates README file with GCC11 and adds more archs (ARM) for GCC11.


uilianries:
I needed to import from ubuntu-toolchain-r, because GCC 11 is not available for Focal (18.04), only for hirsute (21.04), but it's not LTS and it implicates in a future problem.

Issue #136, opened 2019-06-26
[conan 2.0] compatibility with Azure Pipelines and GitHub actions (not using USER)


SSE4:
it would be nice to add compatibility with Azure pipelines
currently, usage of our docker images is impossible, it fails with the following error:
```
##[command]/usr/bin/docker exec  96150cda488b893336a1a922b7400c1d571e0398e449bfc8cbbb3ee9d943fe47 groupadd azure_pipelines_sudo
groupadd: Permission denied.
groupadd: cannot lock /etc/group; try again later.
##[error]Docker exec fail with exit code 10
```
it happens because AZP runs `groupadd` (without sudo), but our containers use user `conan` by default, which has no permissions to run groupadd.
this command is executed by AZP itself and cannot be changed.
two possible solutions:
- switch containers to root user
- add necessary permissions to conan user

/cc @uilianries @ericLemanissier 

### Environment Details
* Conan Docker Tools image: **conanio/gcc7:1.13.1**
* Operating System: **Linux**
* Operation System Version: **Ubuntu 16.04**
* Python version (If you are building some image): **python 3.7.0**

uilianries:
Since our case is very specific I've opened a question on Stack Overflow to receive a feedback from Docker people.

https://stackoverflow.com/questions/56773131/why-not-using-root-user-as-default/

I also tried in Docker forum days ago, but I didn't receive any feedback.

Croydon:
In which setup is this happening? Qt is running partially on AZP https://dev.azure.com/bincrafters/packages/_build?definitionId=7&_a=summary

uilianries:
It's running partially because we have a problem when building, because Qt is huge. It would be nice having Qt running there in AZP to save some time from Travis. However, using "Container Jobs" we need to use the root user because AZP requires extra commands on the entrypoint.

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops

The idea is the same from Gitlab and CircleCI, you just pass the docker image that you want. It's nice because we don't need to run CPT as docker orchestrator.

SSE4:
seems like only root may write to `/etc/group`, and I don't think we should change permissions on it.
```
-rw-r--r-- 1 root root 521 Jun 27 10:05 /etc/group
```
so, I suppose the only way to go is to use root user

uilianries:
Good comment there: https://stackoverflow.com/a/56781394/2036859

[...]

PR #267, opened 2021-06-07
New Docker images: DRY

### LINKED ISSUE #205: Use base docker image for multiple compiler versions

uilianries:
Hello all!

This PR brings a new age for CDT, we no longer will bring new docker recipes for each new compiler release. The README file explains more, but in summary:

- We will keep old Docker recipes and images for a not short period, we understand there are companies and users using them and switch Docker images could be a problem when rebuilding all packages again.
- Those new images will be used as Official, which means, Conan Center will consume them instead the old ones.
- We will to migrate from old to new images, it may take weeks, so don't worry, it won't break you now.
- All new images are based on Ubuntu Xenial (16.04), which is supported until 2024. After that, we can update to 18.04.
- Packages which require GCC 4.9 or older (Ubuntu Trusty) will need to be rebuilt, because glibc version is newer in Xenial.
- We are using the exactly same libstdc++ version for all images, libstdc++.so.6.0.29 which is provided by GCC 11.
- All compilers are built from sources, which give us more compatibility over same environment.
- We had real problems building Clang 9 and older, so we didn't included them. As Clang is retro-compatible, and GCC is still the main compiler used, we believe the last 3 Clang releases are enough.
- All old recipes, scripts and tests were moved to legacy folder, so worry, I didn't change nothing there.
- Special thanks to @SSE4 for helping! 

closes #205

/cc @Croydon 


SSE4:
sorry, but 267 files changed... sounds too much to review in one shot.
you modified lots of existing stuff, e.g. renamed, moved things around - may we avoid it as a first iteration?
I mean, keep existing (old) images with same names, directories and so on, so readme and CI files are mostly unmodified.
and new stuff goes to its own sub-directories and its own entry point.

uilianries:
@SSE4 Agreed, 267 is huge!

I followed your advice, moved to a new folder named "modern". All old stuff is at same place, no changes.

jgsogo:
I really appreciate all the effort here, it's been huge, but very very needed indeed. Now it starts the hard part, with all the reviews and comments and details, the journey has just started 🚀 

On `conan-center-index` side it will take time to start using these images once they are approved and merged. It will require us to force-build a lot of packages but hopefully this will increase the compatibility of the packages generated as they all will share a common ground.

Now we can be sure that any package that is generated using C3i can be used to build any other package. Yes, I'm thinking about build-requires, now **we will be sure that those executables can run on any of our docker images**, we no longer need to worry about `glibc` or `libstdc++` versions... and these are very good news to start thinking about cross-building scenarios where we will need some build-requires.

Now it is our time to spot possible bugs, ask you to change things,... but the work already here is amazing. Thanks again and congratulations, @uilianries (and @SSE4, I know you have talked a lot while developing these changes). 🎉 🤟 🎉 🤟 🎉 🤟

jgsogo (commented):
Some comments to the README

modern/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+Second, we decided to use a single base image, not too old and not the latest, something in the middle, to keep a glibc still compatible for some old distro releases. Thus, we choose Ubuntu 16.04 LTS (Xenial) which its EOL is in 2024. When close to the EOL date, we can move to 18.04 and so on.
+
+As the compiler version is always a problem to align according to the distro release, we decided to build all compilers from sources, which includes GCC and Clang. Although this increased the build time considerably, we still now have full control over the compiler used in our images.
+
+We totally understand that companies and users are still using our “legacy” docker images, so we won’t deprecate them soon. This kind of change requires rebuilding all packages again, and depending on the number, it could take weeks. Thus, don’t worry, we won’t remove them from Docker hub.
+
+**TL;DR** New Docker images will use Ubuntu 16.04 as base and build all compilers from sources. Thus, all binaries generated will link with the same glibc and stdlibc++ versions.
```
prince-chrismc, line 28: Can this be the top of the section?

modern/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Official Docker Images
+
+These are the images uploaded to Docker Hub and currently used by [Conan Center](https://conan.io/center):
+
+#### GCC
+
+| Version                                                                            | Arch    |  Status, Life cycle  |
+|------------------------------------------------------------------------------------|---------|----------------------|
+| [conanio/gcc5-ubuntu16.04: gcc 5](https://hub.docker.com/r/conanio/gcc5-ubuntu16.04/)          | x86_64  |  Supported           |
```
prince-chrismc, line 42: Should the table be updated?

```suggestion
| [conanio/gcc5-ubuntu16.04: gcc 5](https://hub.docker.com/r/conanio/gcc5-ubuntu16.04/)          | x86_64  |  Maintenance Only         |
```

modern/README.md
```diff
@@ -0,0 +1,193 @@
...
+docker run --rm -v /tmp/.conan:/home/conan/.conan conanio/gcc63 bash -c "conan install boost/1.74.0@ --build"
+```
+
+This command is sharing ``/tmp/.conan`` as a shared folder with the conan home, so the Boost package will be built there.
+You can change the directory or execute any other command that works for your needs.
+
+If you are familiarized with Docker compose, also it's possible to start a new container by:
+
+```
+docker-compose run -v /tmp/.conan:/home/conan/.conan gcc11 bash -c "conan install boost/1.74.0@ --build"
```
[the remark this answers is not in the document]
jgsogo: I was asking because of the `gcc11`, only that single set of characters there. If using an image name it works, sure... but, what is `gcc11` referring to?

modern/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+### New Docker Strategy
+
+After many updates, new compiler releases, instability and incompatibility problems, we decided to clean the house, move a step forward with Conan Docker Tools. So far, we usually follow the same recipe, when a compiler version is released, we also release a new docker image, but we have two main problems:
+
+- Ubuntu doesn’t package new compiler versions for each distro release, so we need to use a new Ubuntu version as base image. As a consequence, non LTS versions become a problem when EOL comes.
+- Each Ubuntu release uses different package versions, including very important projects, like glibc, which results in incompatibility from a package created by distro version to another version.
+
+To sanity our problems, we decided to keep focused on the purpose of this project: Docker files for Conan Center CI. So, first we removed older compiler versions which are no longer popular (based on Conan Center download counter).
```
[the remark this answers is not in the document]
uilianries: Done.

modern/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+As the compiler version is always a problem to align according to the distro release, we decided to build all compilers from sources, which includes GCC and Clang. Although this increased the build time considerably, we still now have full control over the compiler used in our images.
+
+We totally understand that companies and users are still using our “legacy” docker images, so we won’t deprecate them soon. This kind of change requires rebuilding all packages again, and depending on the number, it could take weeks. Thus, don’t worry, we won’t remove them from Docker hub.
+
+**TL;DR** New Docker images will use Ubuntu 16.04 as base and build all compilers from sources.
+
+### Legacy Docker Images
+
+If you are looking for legacy docker images (e.g. conanio/gcc10), visit [legacy](legacy) folder.
```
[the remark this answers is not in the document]
uilianries: Done.

modern/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+Second, we decided to use a single base image, not too old and not the latest, something in the middle, to keep a glibc still compatible for some old distro releases. Thus, we choose Ubuntu 16.04 LTS (Xenial) which its EOL is in 2024. When close to the EOL date, we can move to 18.04 and so on.
+
+As the compiler version is always a problem to align according to the distro release, we decided to build all compilers from sources, which includes GCC and Clang. Although this increased the build time considerably, we still now have full control over the compiler used in our images.
+
+We totally understand that companies and users are still using our “legacy” docker images, so we won’t deprecate them soon. This kind of change requires rebuilding all packages again, and depending on the number, it could take weeks. Thus, don’t worry, we won’t remove them from Docker hub.
+
+**TL;DR** New Docker images will use Ubuntu 16.04 as base and build all compilers from sources.
```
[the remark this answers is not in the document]
uilianries: Done.

prince-chrismc (commented):
Considering the rapid shift to cloud, new software being built for Alpine... does the approach you've taken carry through to other "base distros" 

I like the idea of one distro and build for the Linux compatibility 👍

modern/README.md
uilianries: Sorry my ignorance, but what's the difference? I understand "Supported" as "maintained".

modern/README.md
uilianries: Yes, make more sense. I'll update.

uilianries: Done.

modern/README.md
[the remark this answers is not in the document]
jgsogo: Oh! I didn't notice the `docker-compose` file. Thanks!

modern/README.md
jgsogo: For me there is some difference: supported means that we will _add features_ (new compilers, new CMake versions,... things like that) while maintenance means that we will only fix bugs, but no changes will be made otherwise.

However, this table is listing the new images, so I think `Supported` is better than `Maintenance Only` here, but I'm not a native English speaker, so maybe the misunderstanding is on my side.

jgsogo:
> Considering the rapid shift to cloud, new software being built for Alpine... does the approach you've taken carry through to other "base distros"

I'm really a big fan of Alpine and I mentioned it to @SSE4 some days ago, that we could consider building for Alpine as well. Not sure if the approach in CCI would be to cross-compile or use a new collection of docker images based in Alpine. In this repo it would make a lot of sense to provide images for Alpine using the same approach. BUT, let's finish with the Ubuntu-based ones, probably much more needed, and then we can think about Alpine. @prince-chrismc , would you mind open an issue? It is not something we can work on right now, but if someone using this same approach wants to contribute... 😃

modern/README.md
prince-chrismc: You are spot on, I saw "Legacy Docker Images" above the table but missed the following header

 https://github.com/conan-io/conan-docker-tools/pull/267/files#diff-679cc3486654be2732971ec20fa5b84d5348d5dc72f6cca5fe6f102515ba9ac6R31

modern/base/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    && tar -xzf cmake-${CMAKE_VERSION_FULL}-Linux-x86_64.tar.gz \
+       --exclude=bin/cmake-gui \
+       --exclude=doc/cmake \
+       --exclude=share/cmake-${CMAKE_VERSION}/Help \
+       --exclude=share/vim \
+       --exclude=share/vim \
+    && cp -fR cmake-${CMAKE_VERSION_FULL}-Linux-x86_64/* /usr \
+    && rm -rf /root/cmake-${CMAKE_VERSION_FULL}-Linux-x86_64 \
+    && rm /root/cmake-${CMAKE_VERSION_FULL}-Linux-x86_64.tar.gz \
+    && wget -q --no-check-certificate -O /usr/local/bin/jfrog https://api.bintray.com/content/jfrog/jfrog-cli-go/\$latest/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64 \
```
jgsogo, line 50: Change to something official (Bintray is closed):
 * https://jfrog.com/getcli/ , or
 * https://github.com/jfrog/jfrog-cli/releases

modern/base/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    && apt-get -qq autoremove -y \
+    && apt-get -qq autoclean \
+    && apt-get -qq update \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /tmp/* \
+    && runuser -u conan -- printf 'eval "$(pyenv init -)"\n' >> /home/conan/.bashrc \
+    && runuser -u conan -- printf 'eval "$(pyenv virtualenv-init -)"\n' >> ~/.bashrc \
+    && chown -R conan:1001 /opt/pyenv
+
+USER conan
```
jgsogo, line 92: IIRC, we had some issues because we were using `conan` user instead of sudo. Is this what we want? pros and cons?

modern/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+If you are familiarized with Docker compose, also it's possible to start a new container by:
+
+```
+docker-compose run -v /tmp/.conan:/home/conan/.conan gcc11 bash -c "conan install boost/1.74.0@ --build"
+```
+
+
+Build, Test and Deploy
+======================
```
jgsogo, line 118: I would document first how to build (and test) these images using bare docker (or docker-compose) commands, wdyt? It is what I'd expect when reading about docker.

Then, in another section, we can talk about the `run.py` script, which is related to _batch_ generation of images, some shortcuts to trigger one build after the other (it also uses some py files from CPT, which is unexpected if you just want to build one of these images)

modern/README.md
uilianries: So we update the Legacy images to `Maintenance Only`, right?

modern/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Introduce
+The images are already built and uploaded to [conanio](https://hub.docker.com/r/conanio/) dockerhub account, If you want to build your own images you can do it by:
+
+```
+$ GCC_VERSIONS=10 python build.py
```
jgsogo, line 124:
```suggestion
$ GCC_VERSIONS=10 python run.py
```

modern/README.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+The script *build.py* will build, test and deploy your Docker image. You can configure all stages by environment variables listed below.
+
+Also, you can build only a version:
+
+E.g Build and test only a image with Conan and gcc-6
+```
+$ CONAN_GCC_VERSIONS="6" python build.py
```
jgsogo, line 133:
```suggestion
$ CONAN_GCC_VERSIONS="6" python run.py
```

modern/README.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+E.g Build and test only the images with Conan and clang-9, clang-10
+```
+$ CONAN_CLANG_VERSIONS="9,10" python build.py
```
jgsogo, line 138:
```suggestion
$ CONAN_CLANG_VERSIONS="9,10" python run.py
```

modern/base/Dockerfile
uilianries: It's badly recommended running sudo/root user by default due security flaws. I can list some references:
https://cloud.google.com/architecture/best-practices-for-building-containers
https://snyk.io/blog/10-docker-image-security-best-practices/

Unfortunately, 12factor doesn't list security practices, but it's a well known point.

The cons: If you mount a volume which the user is root, the container won't be able to copy directly, you will need to use sudo command.

modern/base/Dockerfile
uilianries: I remember I took this link from the first link, from its script. But I agree, the wrapper is better, if JFrog changes the script, it would be a problem. I'll update it.

modern/README.md
uilianries: That's an excellent addition. I'll add it.

modern/README.md
uilianries: Ops! copy/paste

modern/clang/Dockerfile
```diff
@@ -0,0 +1,149 @@
+ARG DOCKER_USERNAME
+
+FROM ubuntu:xenial as builder
+
+ARG LLVM_VERSION ARTIFACTORY_REPOSITORY LIBSTDCPP_VERSION
+
+LABEL maintainer="Conan.io <info@conan.io>"
+
+# FIXME: Clang 9 and 10 fail to build libcxx because they try to link libc++abi from system.
+#        We could add a patch with target_link_directories, but I prefered copying libc++abi to /usr/lib
```
jgsogo, line 10: Is this comment true for Clang 10? I think we need to update thing `FIXME`.

uilianries: Yes, is true. At line 112 there a copy which fixes it.

modern/base/Dockerfile
uilianries: Done.

modern/README.md
uilianries: Done, I've update README file.

modern/README.md
uilianries: Done

modern/README.md
uilianries: Done

modern/README.md
uilianries: Done

modern/README.md
uilianries: Okay, I'll do it in a next PR, I don't want to touch current files in CDT, we already have 48 changes in this PR, and losing an entire afternoon reviewing a single PR is not cool.

modern/README.md
```diff
@@ -0,0 +1,292 @@
...
+Build, Test and Deploy
+======================
+
+## Introduce
+
+The images are already built and uploaded to [conanio](https://hub.docker.com/r/conanio/) dockerhub account, but if you want to build your own images, read this section.
+
+The `docker-compose.yml` file list all possible combinations of images and it will be explained below:
+
+### Build
```
jgsogo, line 126: Before building `clang12` as in the example below, there is one more stage: building `base` image. I would add here a comment, maybe "Pre-build" stage saying that if you want to customize the version of Conan, CMake or Python, you need to edit the `.env` file and generate again the `base` image.

modern/gcc/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+       libmpc-dev \
+       nasm \
+       dh-autoreconf \
+       libffi-dev \
+       libssl-dev \
+       pkg-config \
+    && wget -q --no-check-certificate http://mirrors.concertpass.com/gcc/releases/gcc-${GCC_VERSION}/gcc-${GCC_VERSION}.tar.xz \
+    && tar Jxf gcc-${GCC_VERSION}.tar.xz \
+    && cd gcc-${GCC_VERSION} \
+    && ./configure --build=x86_64-linux-gnu \
```
SSE4, line 30: may we download pre-built compiler if available from artifactory generic repository here?
I don't want to waste 1 hour * number of compilers each time I modify the image
(e.g. bump `6.0.29` -> `6.0.30`)
I think rebuild of the compiler is really rarely (RARELY) required, only if we modify e.g. configure flags for the compiler itself.

modern/gcc/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+RUN sudo rm -rf /usr/lib/gcc/x86_64-linux-gnu/* \
+    && sudo cp -a /tmp/install/lib/gcc/x86_64-linux-gnu/${GCC_VERSION} /usr/lib/gcc/x86_64-linux-gnu/ \
+    && sudo cp -a /tmp/install/include/* /usr/local/include/ \
+    && sudo cp -a /tmp/install/lib64/ /usr/local/ \
+    && sudo cp -a /tmp/install/libexec/ /usr/local/ \
+    && sudo cp -a /tmp/install/lib/* /usr/local/lib/ \
+    && sudo cp -a /tmp/install/bin/* /usr/local/bin/ \
+    && sudo rm -rf /tmp/install \
+    && sudo mv /tmp/gcc/lib64/libstdc++.so.6.0.29 /usr/local/lib64/libstdc++.so.6.0.29 \
```
SSE4, line 60: can we have `6.0.29` as a variable, so I don't need to change the number in multiple places?

modern/base/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+    && pip install -q --no-cache-dir conan==${CONAN_VERSION} conan-package-tools \
+    && printf '/usr/local/lib\n' > /etc/ld.so.conf.d/20local-lib.conf \
+    && printf '/usr/local/lib64\n' >> /etc/ld.so.conf.d/20local-lib.conf \
+    && mv /etc/ld.so.conf.d/libc.conf /etc/ld.so.conf.d/10libc.conf \
+    && mv /etc/ld.so.conf.d/x86_64-linux-gnu.conf /etc/ld.so.conf.d/30x86_64-linux-gnu.conf \
+    && mv /etc/ld.so.conf.d/zz_i386-biarch-compat.conf /etc/ld.so.conf.d/40zz_i386-biarch-compat.conf \
+    && mv /etc/ld.so.conf.d/zz_x32-biarch-compat.conf /etc/ld.so.conf.d/50zz_x32-biarch-compat.conf \
+    && rm /etc/ld.so.cache \
+    && ldconfig -C /etc/ld.so.cache \
+    && apt-get -qq purge -y g++-multilib gcc gcc-5 \
```
jgsogo, line 83: `g++-multilib` is installed and purged. Is it on purpose?

uilianries: Yes. Pyenv build Python from source, so we need a compiler, otherwise, I would remove it. I think pyenv is still the best python installer for our case, because it has been working well, but I'm totally open for new suggestions

modern/clang/Dockerfile
uilianries: I'll keep Clang 9 listed as info. As you can see, each llvm version has a different behavior.

modern/README.md
uilianries: Good addition, I'll update it.

modern/gcc/Dockerfile
uilianries: I prefer keep the recipe "how to build from sources", as an enterprise, I would not like to download a magic package with binaries. The builder stage is isolated from others, is not something to be rebuilt every time, you build once, cache it and only rebuild the Deploy step.

modern/gcc/Dockerfile
uilianries: Yes, totally. I'll update it.

modern/README.md
uilianries: Done.

modern/README.md
```diff
@@ -131,21 +131,59 @@ The `DOCKER_USERNAME` is an environment variable not filled by the `.env` file,
...
 image, easier to be maintained and isolated from the environment used to build the compiler.
 
+This image avoids all Builder cache, using Docker multi-stage feature, we copy only the compiler installation folder.
+
+It's the default build command, for instance:
+
+    $ export DOCKER_USERNAME=conanio
+    $ docker-compose build gcc10
+
+#### Update stage (Conan)
```
jgsogo, line 176: I don't find this stage useful and I think it introduces something outside workflow (this was my intention here: https://github.com/uilianries/conan-docker-tools/pull/4).

If you want to change the Conan version there is no need for a _massive_ build, you just need to build `base` again and then the `deploy` stage, there is no need for shortcuts, no need to have one additional image name (these are using a `-conan` word in the middle). TBH, I would remove these `-conan` images.

jgsogo:
https://github.com/conan-io/conan-docker-tools/pull/267

azure-pipelines.yml
```diff
@@ -1,7 +1,54 @@
 jobs:
-- job: Docker
+- job: Modern
```
SSE4, line 2: to be honest, I don't like naming like `Modern`.
okay, today it's modern for sure, but in a year or several it may easily became outdated, and will only introduce a confusion.
I'd say let's use some non-timing naming, such as `same-glibc`, `xenial-based`, or whatever.

modern/base/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    && tar -xzf cmake-${CMAKE_VERSION_FULL}-Linux-x86_64.tar.gz \
+       --exclude=bin/cmake-gui \
+       --exclude=doc/cmake \
+       --exclude=share/cmake-${CMAKE_VERSION}/Help \
+       --exclude=share/vim \
+       --exclude=share/vim \
+    && cp -fR cmake-${CMAKE_VERSION_FULL}-Linux-x86_64/* /usr \
+    && rm -rf /root/cmake-${CMAKE_VERSION_FULL}-Linux-x86_64 \
+    && rm /root/cmake-${CMAKE_VERSION_FULL}-Linux-x86_64.tar.gz \
+    && curl -fL https://getcli.jfrog.io | sh \
```
SSE4, line 50: this line doesn't seem to be reproducible, can we download a specific version or check sha?

modern/base/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    && chmod +x /usr/local/bin/jfrog \
+    && groupadd 1001 -g 1001 \
+    && groupadd 1000 -g 1000 \
+    && groupadd 2000 -g 2000 \
+    && groupadd 999 -g 999 \
+    && useradd -ms /bin/bash conan -g 1001 -G 1000,2000,999 \
+    && printf "conan:conan" | chpasswd \
+    && adduser conan sudo \
+    && printf "conan ALL= NOPASSWD: ALL\\n" >> /etc/sudoers \
+    && wget --no-check-certificate --quiet -O /tmp/pyenv-installer https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer \
```
SSE4, line 61: same, downloading from master, I'd prefer a specific release to avoid sudden breakage, if possible

modern/base/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    && curl -fL https://getcli.jfrog.io | sh \
+    && mv jfrog /usr/local/bin/jfrog \
+    && chmod +x /usr/local/bin/jfrog \
+    && groupadd 1001 -g 1001 \
+    && groupadd 1000 -g 1000 \
+    && groupadd 2000 -g 2000 \
+    && groupadd 999 -g 999 \
+    && useradd -ms /bin/bash conan -g 1001 -G 1000,2000,999 \
+    && printf "conan:conan" | chpasswd \
+    && adduser conan sudo \
```
SSE4, line 59: we already know conan user is problematic for CI servies: #136
should we use it for new images? why it might be required?

modern/base/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    && update-alternatives --install /usr/bin/python3 python3 /opt/pyenv/shims/python3 100 \
+    && update-alternatives --install /usr/bin/pip3 pip3 /opt/pyenv/shims/pip3 100 \
+    && update-alternatives --install /usr/local/bin/python python /opt/pyenv/shims/python 100 \
+    && update-alternatives --install /usr/local/bin/pip pip /opt/pyenv/shims/pip 100 \
+    && pip install -q --upgrade --no-cache-dir pip \
+    && printf '/usr/local/lib\n' > /etc/ld.so.conf.d/20local-lib.conf \
+    && printf '/usr/local/lib64\n' >> /etc/ld.so.conf.d/20local-lib.conf \
+    && mv /etc/ld.so.conf.d/libc.conf /etc/ld.so.conf.d/10libc.conf \
+    && mv /etc/ld.so.conf.d/x86_64-linux-gnu.conf /etc/ld.so.conf.d/30x86_64-linux-gnu.conf \
+    && mv /etc/ld.so.conf.d/zz_i386-biarch-compat.conf /etc/ld.so.conf.d/40zz_i386-biarch-compat.conf \
```
SSE4, line 78: what does `zz_` prefix mean?

modern/clang/Dockerfile
```diff
@@ -0,0 +1,154 @@
+ARG DOCKER_USERNAME
```
SSE4, line 1: so, this is not an actual `clang` image, but instead, it's just an intermediate builder image, which we don't expect to be used by consumers, correct?
maybe then rename it to the `clang-builder` to avoid confusion? it might not be so clear which Dockerfile do I need to modify, especially for newcomers

modern/server/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM python:3.8-alpine3.8
```
SSE4, line 1: is this server image different from the legacy one?
for server, we don't need any compilers and other stuff, just python and python packages, so I don't see any reason why should we modify it right now

azure-pipelines.yml
uilianries: Done. Renamed to "Xenial".

modern/base/Dockerfile
uilianries: Done. Using a commit now. The latest tag is from 2015.

modern/base/Dockerfile
uilianries: Done. Now it will used the latest version available, which is 1.48.1

modern/base/Dockerfile
uilianries: It's a problem related to Azure, but we I prefer following security recommendations than allowing Azure. If necessary, we can move to Jenkins.

There some good blog posts about it:
cloud.google.com/architecture/best-practices-for-building-containers
snyk.io/blog/10-docker-image-security-best-practices

modern/base/Dockerfile
uilianries: It's something from Debian, it represents patch version, like x.y.zz.

Here, I've added 40 as prefix, because ldconfig used alphabetical order to load all files.

modern/clang/Dockerfile
uilianries: Correct, it's an intermediate layer which builds the compiler. Okay, clang-builder sounds a good name.

modern/server/Dockerfile
uilianries: Agreed. Done, removed.

modern/base/Dockerfile
uilianries: Looking again, numbers come first than letters, so we don't need it, I'll keep the default.

jgsogo:
RFC:
 * I think the base image should have Conan installed, and use the Conan version as a tag (it is consistent with the rest of the images). No need for those `-conan` images and there is no very little extra time when releasing images with new Conan, the `-deploy` image is just copying things, it doesn't need to build anything again.
 

 * By default _preconfigure_ Conan client and update `default` profile with the _new_  `compiler.libcxx` configuration:
 
    ``` conan profile update settings.compiler.libcxx=libstdc++11 default```

uilianries:
> I think the base image should have Conan installed, and use the Conan version as a tag (it is consistent with the rest of the images). No need for those -conan images and there is no very little extra time when releasing images with new Conan, the -deploy image is just copying things, it doesn't need to build anything again.

What is your suggestion about new Conan releases, build base and deploy again for each new release?
Remember that if we need to update anything in base right now, we would need re-execute all steps again.
Of course, we can separate `RUN pip install conan...` in base, it would increase a little bit the image size, but the trade off is still valid, because we are doing it on top layer (conan). I like your idea, at least, we would have all centralized.

> By default preconfigure Conan client and update default profile with the new compiler.libcxx configuration

I would prefer keeping the old format for compatibility reason. People still can use our Docker images as build environment, but won't be the same default behavior as when installing locally. Example, you build a package in your local machine, then you try to consume it in your docker image, the libcxx is not same, so Conan complains about and you have to decide between building all again or updating the default profile.

jgsogo:
> > By default preconfigure Conan client and update default profile with the new compiler.libcxx configuration
> 
> I would prefer keeping the old format for compatibility reason. People still can use our Docker images as build environment, but won't be the same default behavior as when installing locally. Example, you build a package in your local machine, then you try to consume it in your docker image, the libcxx is not same, so Conan complains about and you have to decide between building all again or updating the default profile.

Agree with you on the defaults and compatibility reasons... but with Conan v2 in mind, `libstdc++11` will be the default. It is also the default for new compiler. This cannot be considered as breaking because these are new images with different names, we need to think about them on the long-term, not constraint to current images or technical debt.

uilianries:
Closing this PR in favor of #267

The new PR is the **SAME**, but uses CDT as repository, so any new PR  will be tested on Azure.

PR #271, opened 2021-06-16
Modern Docker images: Base - Builder - Deploy

### LINKED ISSUE #205: Use base docker image for multiple compiler versions

uilianries:
### NOTE: This PR replaces #267. So any new contribution can use Azure for testing.

Hello all!

This PR brings a new age for CDT, we no longer will bring new docker recipes for each new compiler release. The README file explains more, but in summary:

- We will keep old Docker recipes and images for a not short period, we understand there are companies and users using them and switch Docker images could be a problem when rebuilding all packages again.
- Those new images will be used as Official, which means, Conan Center will consume them instead the old ones.
- We will to migrate from old to new images, it may take weeks, so don't worry, it won't break you now.
- All new images are based on Ubuntu Xenial (16.04), which is supported until 2024. After that, we can update to 18.04.
- Packages which require GCC 4.9 or older (Ubuntu Trusty) will need to be rebuilt, because glibc version is newer in Xenial.
- We are using the exactly same libstdc++ version for all images, libstdc++.so.6.0.29 which is provided by GCC 11.
- All compilers are built from sources, which give us more compatibility over same environment.
- We had real problems building Clang 9 and older, so we didn't included them. As Clang is retro-compatible, and GCC is still the main compiler used, we believe the last 3 Clang releases are enough.
- All old recipes, scripts and tests were moved to legacy folder, so worry, I didn't change nothing there.
- Special thanks to @SSE4 and @jgsogo for helping! 

closes #205

/cc @Croydon 


uilianries:
> > > By default preconfigure Conan client and update default profile with the new compiler.libcxx configuration
> >
> > I would prefer keeping the old format for compatibility reason. People still can use our Docker images as build environment, but won't be the same default behavior as when installing locally. Example, you build a package in your local machine, then you try to consume it in your docker image, the libcxx is not same, so Conan complains about and you have to decide between building all again or updating the default profile.
>
> Agree with you on the defaults and compatibility reasons... but with Conan v2 in mind, `libstdc++11` will be the default. It is also the default for new compiler. This cannot be considered as breaking because these are new images with different names, we need to think about them on the long-term, not constraint to current images or technical debt.

@jgsogo Indeed we have to consider Conan 2.x, but we can update libcxx according the client version. It's something easy to be implemented.

jgsogo:
Easy failure on branch (almost expected):

```
groovy.lang.MissingPropertyException: No such property: CHANGE_TARGET for class: groovy.lang.Binding
```

uilianries:
> Easy failure on branch (almost expected):
> 
> ```
> groovy.lang.MissingPropertyException: No such property: CHANGE_TARGET for class: groovy.lang.Binding
> ```

Can it be fixed or will be fixed after merging to master branch?

jgsogo:
It needs to be fixed. That variables doesn't exist in branches, only in PRs

jgsogo:
I have no idea why the CI fails in the `branch` 🤷   I'm afraid it has something to do with concurrency issues when uploading/downloading docker images with the repository (branch and PR use some same images).

I know it is not a fix, but I've configured Jenkins to run only the branches that are not filed as PRs.

uilianries:
@jgsogo @SSE4 I think we are good to merge now. Let's do a last review.

jgsogo:
Please, consider https://github.com/conan-io/conan-docker-tools/pull/291

uilianries:
> Please, consider #291

Merged

jgsogo (approved):
IMO, this is ready to go. We can improve it after merging to `master` (upload to docker hub).

PR #274, opened 2021-06-16
Propose new test framework for the docker images

jgsogo:
Getting some bits from here: https://github.com/uilianries/conan-docker-tools/pull/5

---

This is the proposal in this PR (although it doesn't modify anything already existing):
 * Get rid of all the Python code to generate docker images (remove dependency of `conan` and `conan-package-tools`). Use plain and well-known docker commands and mechanisms to generate the images. Users should be able to understand this repository and generate the images without the need of inspecting/installing any `.py` file or Python itself.
 
 * Replace current testing with other that uses only the `pytest` framework to orchestrate the commands. Get rid of `.sh` scripts and `run.py` scripts to run the tests and the checks. New test framework will operate on a given image name:
 
    ```
    pytest modern/tests --image docker.io/conanio/gcc10-ubuntu16.04:1.37.2
    ```

    Understanding the test framework requires Python knowledge, but IMHO here there is a huge advantage in using Python over plain `.sh` scripts: better reporting, clear checks,...
    
---

If we agree on this, next PRs will start to actually migrate tests from the old _framework_ to the new one. Eventually, all tests will be migrated (and more added) and we can remove legacy and adapt README files and so on.

wdyt, @SSE4 @uilianries ?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,10 +10,10 @@ jobs:
       architecture: 'x64'
   - script: |
       pip install -U docker-compose conan conan-package-tools
-      cd modern && python run.py
+      cd modern
+      python run.py
       pip uninstall --yes conan conan-package-tools
-      ls -la
-      cd modern && ./run_pytest.sh
+      ./run_pytest.sh
     env:
       DOCKER_LOGIN_USERNAME: $(DOCKER_LOGIN_USERNAME)
       DOCKER_PASSWORD: $(DOCKER_PASSWORD)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,6 +11,8 @@ jobs:
   - script: |
       pip install -U docker-compose conan conan-package-tools
       cd modern && python run.py
+      pip uninstall --yes conan conan-package-tools
+      ./modern/run_pytest.sh
     env:
       DOCKER_LOGIN_USERNAME: $(DOCKER_LOGIN_USERNAME)
       DOCKER_PASSWORD: $(DOCKER_PASSWORD)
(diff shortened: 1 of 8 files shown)


azure-pipelines.yml
```diff
@@ -11,40 +11,52 @@ jobs:
   - script: |
       pip install -U docker-compose conan conan-package-tools
       cd modern && python run.py
+      pip uninstall --yes conan conan-package-tools
+      ./modern/run_pytest.sh
```
jgsogo, line 15: Let's get rid of the logic hidden in Python scripts, too much magic. This `sh` could be copy/pasted here if needed. Just use plain commands that are easier to understand.

azure-pipelines.yml
```diff
@@ -11,40 +11,52 @@ jobs:
...
     env:
       DOCKER_LOGIN_USERNAME: $(DOCKER_LOGIN_USERNAME)
       DOCKER_PASSWORD: $(DOCKER_PASSWORD)
   strategy:
     matrix:
       Ubuntu 16.04 GCC 5 x86_64:
         GCC_VERSIONS: "5"
+        GCC_VERSION: "5"
```
jgsogo, line 23: The CI won't build several images in the same matrix entry. Users should prefer known docker commands over magic hidden in python scripts. If they want to build several images, they can just concatenate several commands. Avoid complexity on our side.

uilianries: Okay, let keep this PR more focused on tests. I can do it directly in single-image branch. It also required README update.

jgsogo:
It works! 🎉 

Output from the testing part:

```
Docker username: conanio
Conan version: 1.37.2
GCC_VERSION 5
CLANG_VERSION
docker image: docker.io/conanio/gcc5-ubuntu16.04:1.37.2
============================= test session starts ==============================
platform linux -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/vsts/work/1/s/modern
collected 1 item

tests/test_versions/test_installed.py .                                  [100%]

============================== 1 passed in 0.85s ===============================
```

uilianries:
Working here:

```
```
$ pytest -v modern/tests --image docker.io/conanio/gcc10-ubuntu16.04:1.37.2
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.8.8, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /opt/pyenv/versions/3.8.8/bin/python3.8
cachedir: .pytest_cache
rootdir: /home/uilian/Development/conan/conan-docker-tools
plugins: forked-1.3.0, xdist-2.2.1
collected 1 item                                                                                                                                                                                                                             

modern/tests/test_versions/test_installed.py::test_cmake_version PASSED

uilianries (approved):
LGTM

PR #275, opened 2021-06-18
[modern] Add very basic tests

jgsogo:
Add very basic tests, check that the tools are installed and versions match

PR #276, opened 2021-06-18
[modern] Add tests for gfortran compiler

jgsogo:
After https://github.com/conan-io/conan-docker-tools/pull/275


modern/tests/test_build/test_gfortran.py
```diff
@@ -0,0 +1,56 @@
...
+            print(out)
+            print(err)
+
+    @pytest.mark.service('deploy', 'jenkins')
+    def test_vanilla_image(self, container, expected):
+        # C executable should run in vanilla image
+        with container.run_container(expected.vanilla_image()) as vanilla:
+            with vanilla.working_dir(self._build_directory):
+                out, err = vanilla.exec(['./hello'])
+                assert 'Hello world!' in out, f"out: '{out}' err: '{err}'"
```
uilianries, line 43:
```suggestion
        out, err = container.exec(['ls', '-Ral', '/usr/local'])
        container.exec(['cp', '/usr/local/lib64/libgfortran.so.3.0.0', self._build_directory])
        container.exec(['cp', '/usr/local/lib64/libquadmath.so.0.0.0', self._build_directory])
        with container.run_container(expected.vanilla_image()) as vanilla:
            with vanilla.working_dir(self._build_directory):
                vanilla.exec(['cp', 'libgfortran.so.3.0.0', '/usr/lib/libgfortran.so.3'])
                vanilla.exec(['cp', 'libquadmath.so.0.0.0', '/usr/lib/libquadmath.so.0'])
                out, err = vanilla.exec(['./hello'])
                assert 'Hello world!' in out, f"out: '{out}' err: '{err}'"
```

We need to copy libgfortran and libquadmath from container. This suggestion is hardcoded and works for GCC5, but I guess the library version is not the same for gcc11 for instance. Still, we can use wildcard for copying.

Can we solve it using apt-get only?
No, because libgfortran is under gcc apt package
Yes, because we can download any .deb package and extract its content.

jgsogo: Ok, if this doesn't work in a _vanilla_ image, probably we shouldn't run this test.

.ci/xenial.jenkinsfile
```diff
@@ -28,14 +28,20 @@ def buildImage(String latestMaster, String latestBranch, String expectedImage, S
...
     String output = sh(script: 'docker ps --no-trunc', returnStdout: true)
-    String line = output.split('\n').each({ it.trim() }).findAll({ it.contains(nodeName) && it.contains('/opt/entrypoint.sh') })[0]
-    String nodeId = line.split(' ')[0]
+    String nodeId = null
+    output.split('\n').each({
+        String line = it.trim()
+        if (line.contains(nodeName) && line.contains('/opt/entrypoint.sh')) {
+            nodeId = line.split(' ')[0]
+        }
+    })
```
jgsogo, line 44: `findAll` returns an iterator, which is not serializable. Depending on where the source is executed, this can lead to SerializationErrors in Jenkins.

jgsogo:
It also fails for regular GCC images:

```

________________ TestBuildGFortran.test_compatible[gcc-10.3.0] _________________



self = <test_build.test_gfortran.TestBuildGFortran object at 0x7feb021d6fd0>

container = <utils.docker_container.DockerContainer object at 0x7feb021b5850>

expected = Expected(distro=Distro(name='ubuntu', version=Version(full_version='16.04', major='16', minor='04', patch=None)), dock...rsion='11.1.0', major='11', minor='1', patch='0'), Version(full_version='12.0.0', major='12', minor='0', patch='0')]}))

compiler = 'gcc', compiler_version = '10.3.0'



    @pytest.mark.service('xtest')

    @pytest.mark.parametrize("compiler, compiler_version", [(key, str(v)) for key, value in get_compiler_versions().items() for v in value])

    def test_compatible(self, container, expected, compiler, compiler_version):

        # Fortran executable requires GCC (gfortran) installed

        if compiler != "gcc":

            pytest.skip('Requires GCC installed')

    

        image_name = expected.image_name(compiler, Version(compiler_version))

        with container.run_container(image_name) as image:

            with image.working_dir(self._build_directory):

                out, err = image.exec(['./hello'])

>               assert 'Hello world!' in out, f"out: '{out}' err: '{err}'"

E               AssertionError: out: '' err: './hello: error while loading shared libraries: libgfortran.so.3: cannot open shared object file: No such file or directory

E                 '

E               assert 'Hello world!' in ''
```

.ci/xenial.jenkinsfile
uilianries: Good to know!

uilianries (approved):
Excellent job! :clap:

PR #277, opened 2021-06-18
[modern] Add tests for the jenkins agent

jgsogo:
After #275


modern/tests/fixtures/expected.py
```diff
@@ -73,6 +74,16 @@ def get_compiler_version(compiler_name, compiler_major):
...
 
 
+def get_jenkins_agent_version():
+    docker_file = os.path.realpath(os.path.join(os.path.dirname(__file__), '..', '..', 'docker-compose.yml'))
+    with open(docker_file, 'r') as f:
+        data = yaml.safe_load(f)
+
+    ss = data.get('x-common-args').get('JENKINS_AGENT_VERSION')
+    m = re.match(r'\${JENKINS_AGENT_VERSION:-([\d.]+)}', ss)
+    return m[1]
```
uilianries, line 84:
```suggestion
    docker_file = os.path.realpath(os.path.join(os.path.dirname(__file__), '..', '..', 'docker-compose.yml'))
    with open(docker_file, 'r') as f:
        data = yaml.safe_load(f)

    ss = data.get('x-common-args').get('JENKINS_AGENT_VERSION')
    m = re.match(r'\${JENKINS_AGENT_VERSION:-([\d.]+)}', ss)
    return m[1]
```

Should we move it to .env to keep all centralized?

PR #280, opened 2021-06-23
[ci] Use Jenkins

jgsogo:
Trying Jenkins approach, first draft. Something like this:

image


modern/.env
```diff
@@ -1,4 +1,5 @@
 CONAN_VERSION=1.37.2
+DOCKER_TAG=1.37.2
```
uilianries, line 2: Why separate DOCKER_TAG and CONAN_VERSION?

.ci/xenial.jenkinsfile
```diff
@@ -0,0 +1,148 @@
+// These are things to parameterize/read/write from/to .env file
+List gccVersions = ['5', '6', '7', '8', '9', '10', '11']
+List clangVersions = ['10', '11', '12']
+boolean isMaster = (scm.branches[0].name == 'master')
+echo ">> scm.branches[0].name: ${scm.branches[0].name}"
+List<String> environment = []
+
+def configure_docker() {
+    withCredentials([usernamePassword(credentialsId: 'c3istg-dad-docker', usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) {
```
uilianries, line 9:
```suggestion
    withCredentials([usernamePassword(credentialsId: 'c3istg-dad-docker', usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) {
```

What about the official hub.docker registry? People will download from there, at least, we could upload a copye. I know using our side could save CI time as all is at same server, but still don't want to configure a new registry only to download a single image.

jgsogo: Sure, PRs should be cached on our side, when the PR is merged to `master` then those images will be built again (or just promoted) to DockerHub.

modern/.env
jgsogo: I need to tag the images built from PRs with the corresponding commit, so they are not used by other jobs running other branches/pull requests. Only images generated in `master` can use the tag with the Conan version.

jgsogo:
Green and building all the images. Total time 1h 41min without using cached images.

uilianries (approved):
LGTM

PR #282, opened 2021-06-28
Move parameters to `.env` file

jgsogo:
After #280


uilianries:
Screenshot 2021-06-28 at 14-52-07 Move parameters to ` env` file by jgsogo · Pull Request #282 · conan-io conan-docker-tools

Amazing!

.ci/xenial.jenkinsfile
```diff
@@ -166,9 +186,26 @@ node('Linux') {
...
+            }
+            clangVersions.each {
+                images.add("clang${versionMajor}-${distro}")
+                images.add("clang${versionMajor}-${distro}-jenkins")
+            }
+
+            images.each { image ->
+                String uploadImage = "${originalUsername}/${image}:${conanVersion}"
+                sh "docker tag ${dockerUsername}/${image}:${branchName}-${commit} ${uploadImage}"
+                // sh "docker push ${uploadImage}"
```
jgsogo, line 206: I want to check this works (image names are properly set) before activating the `push` (see also that the conditional `isMaster` is commented out).

.ci/xenial.jenkinsfile
```diff
@@ -56,6 +66,16 @@ node('Linux') {
...
+
+        // Parse more information
+        distro = job_environment['DISTRO']
+        conanVersion = job_environment['CONAN_VERSION']
+        originalUsername = originalEnvironment['DOCKER_USERNAME']
+
+        echo "gccVersions: ${gccVersions.join(', ')}"
+        echo "clangVersions: ${clangVersions.join(', ')}"
+        assert gccVersions.size() > 0: "Expecting to build some GCC versions at least"
+        assert clangVersions.size() > 0: "Expecting to build some Clang versions at least"
```
uilianries, line 78: What does happen if we want to build only gcc or clang?

modern/base/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM ubuntu:xenial
+ARG DISTRO_VERSION
+
+FROM ubuntu:${DISTRO_VERSION}
```
uilianries, line 3: xenial and 16.04 are tags, but same image. I'm fine use version instead of alias, because is easier to understand the distro age.

.ci/xenial.jenkinsfile
```diff
@@ -166,9 +186,30 @@ node('Linux') {
...
+            }
+            clangVersions.each { v ->
+                String versionMajor = getVersionMajor(v)
+                images.add("clang${versionMajor}-${distro}")
+                images.add("clang${versionMajor}-${distro}-jenkins")
+            }
+
+            images.each { image ->
+                String builtImage = "${dockerUsername}/${image}:${branchName}-${commit}"
+                String uploadImage = "${originalUsername}/${image}:${conanVersion}"
```
jgsogo, line 207: Please, @uilianries , double-check this is the expected final name for the images we are building

.ci/xenial.jenkinsfile
jgsogo: Currently, it is not possible, the CI will always build both of them. If we remove GCC (or Clang) from the repo, then we should remove them from the CI as well.

.ci/xenial.jenkinsfile
uilianries: I would consider using `git rev-parse --short HEAD` at line 41. It's very rare a collision, and is more readable

.ci/xenial.jenkinsfile
jgsogo: TBH, I'm thinking about creating a generic job to build all the images and call it from the multibranch one... it will allow us to manually configure the images and build them (everything could be configurable: conan, python, cmake,... gcc-versions, clang-versions,...), but it is something for the future.

.ci/xenial.jenkinsfile
jgsogo: 👍  np

jgsogo: Anyway, take into account that images tagged that way are not going to be used by humans, they are only used to save the state so we can hydrate the cache afterward.

uilianries:
Screenshot 2021-06-28 at 17-47-53 jenkins ConanDockerTools Xenial PR-282 #6

Image names looks correct.

uilianries (approved):

PR #292, opened 2021-07-14
Validate installed and not installed tools for new Docker images

### LINKED ISSUE #287: [xenial] Test installed tools and applications
Under this directory https://github.com/conan-io/conan-docker-tools/tree/feature/single-image/modern/tests/test_installed we should check that are the required tools are installed (and we might check that some are not installed... for example, if we don't want `pkg_config` to be available and it should be a build-requires).

uilianries:
Some tools like `pkg-config` can not be removed, because they are largely used in Conan Center, if we remove it, first we need to update ALL recipes in CCI, otherwise, those new images will break many users.

Some tools like vim, used to be installed in the past, for "development" when creating a new recipe. At some point we removed it, and we don't want it again.

closes #287

### DIFF
--- a/modern/tests/fixtures/expected.py
+++ b/modern/tests/fixtures/expected.py
@@ -61,9 +61,20 @@ class Expected:
     docker_tag: str
     python: Version
     cmake: Version
+    pkg_config: Version
+    make: Version
+    autoconf: Version
+    perl: Version
+    wget: Version
+    curl: Version
+    git: Version
+    subversion: Version
+    xzutils: Version
+    nasm: Version
     conan: Version = None
     compiler: Compiler = None
     compiler_versions: defaultdict(list) = None
+
 
     def __str__(self):
         return f"""
@@ -76,6 +87,16 @@ class Expected:
             - conan: {self.conan}
             - compiler: {self.compiler}
             - compiler_versions: {self.compiler_versions}
+            - pkg_config: {self.pkg_config}
+            - make: {self.make}
+            - autoconf: {self.autoconf}
+            - perl: {self.perl}
+            - wget: {self.wget}
+            - curl: {self.curl}
+            - git: {self.git}
+            - subversion: {self.subversion}
+            - xzutils: {self.xzutils}
+            - nasm: {self.nasm}
         """
 
     def vanilla_image(self):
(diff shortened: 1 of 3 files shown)


modern/tests/test_installed/test_not_installed.py
```diff
@@ -0,0 +1,24 @@
...
+        "bazel",
+        "scons",
+        "qmake",
+        "xmake",
+        "premake5",
+        "b2",
+        "lua",
+    ],
+)
+def test_not_installed(container, tool):
```
jgsogo, line 21: 👍

modern/tests/fixtures/expected.py
```diff
@@ -132,7 +153,20 @@ def expected(request) -> Expected:
...
+    pkg_config = Version("0.29.1")
+    make = Version("4.1")
+    autoconf = Version("2.69")
+    perl = Version("5.22.1")
+    wget = Version("1.17.1")
+    curl = Version("7.47.0")
+    git = Version("2.7.4")
+    subversion = Version("1.9.3")
+    xzutils = Version("5.1.0alpha")
+    nasm = Version("2.11.08")
```
jgsogo, line 165: These versions depend on the distro version, right? They are not hardcoded in the Dockerfiles, so they might change. They can even change if the version is modified in the _apt repository_.

Should we hardcode the versions in the Dockerfile? Should we map these versions to distro-version in some external file?

jgsogo (commented):
Some thoughts

modern/tests/fixtures/expected.py
uilianries: We can collect those version online actually. But a local file sounds great too.

modern/tests/test_installed/test_installed.py
```diff
@@ -1,9 +1,35 @@
...
+    "git": {"16.04": "2.7.4"},
+    "svn": {"16.04": "1.9.3"},
+    "xz": {"16.04": "5.1.0"},
+    "nasm": {"16.04": "2.11.08"},
+}
+
+
 def test_cmake_version(container, expected):
-    output, _ = container.exec(['cmake', '--version'])
+    output, _ = container.exec(["cmake", "--version"])
```
uilianries, line 20: sorry, python Black

modern/tests/fixtures/expected.py
uilianries: Thinking again, usually Ubuntu delays package releases, and most of time, they don't change the version, only package version. Let's use a local list, if it starts to hurt, we can adopt a dynamic list from internet.

jgsogo (approved):
I'm ok with it because I think all of these are common tools in any developer machine... but I know there are other people thinking that we should use Conan packages (`build-requires`) to retrieve some of these tools. IMHO it is important to save some bandwitdh.

uilianries:
> but I know there are other people thinking that we should use Conan packages (build-requires) to retrieve some of these tools.

Yes, for instance, you will find some recipes in CCI using autotools package. I also agree with them, I prefer Conan packages to keep the reproducibility.

PR #293, opened 2021-07-14
Validate linked libs to c and c++ apps

### LINKED ISSUE #286: [xenial] Test system libraries versions
https://github.com/conan-io/conan-docker-tools/blob/e44f0e8d64dabdd3fabcf3f0ee68b632428ca62a/modern/tests/test_build/test_simple.py#L32

We need to check (following `ldd` and symlinks) that the libraries and versions linked to the executable are the ones we are expecting to link.

uilianries:
closes #286

### DIFF
--- a/modern/tests/fixtures/expected.py
+++ b/modern/tests/fixtures/expected.py
@@ -52,8 +52,6 @@ class Distro:
 class Compiler:
     name: str
     version: Version
-    system_libs_c: list
-    system_libs_cpp: list
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


modern/tests/fixtures/expected.py
```diff
@@ -114,6 +116,29 @@ def get_envfile_values():
...
+                'libstdc++.so.6 => /usr/local/lib64/libstdc++.so.6',
+                'libgcc_s.so.1 => /usr/local/lib64/libgcc_s.so.1',
+        ] + get_system_libs_c()
+    else:
+        return [
+            'libc++abi.so.1 => /usr/local/lib/libc++abi.so.1',
+            'libc++.so.1 => /usr/local/lib/libc++.so.1',
+            'libstdc++.so.6 => /usr/local/lib64/libstdc++.so.6',
+            'libllvm-unwind.so.1 => /usr/local/lib/libllvm-unwind.so.1',
+        ] + get_system_libs_c()
```
jgsogo, line 139: I would try to avoid hardcoded values here. I know it is a `fixture` and sometimes they contain hardcoded values (data to populate testing databases,...) but in this case, everything here is parametrized. 

I would really move these lines to the only test where they are used. wdyt?

uilianries: fair enough

uilianries: Done

jgsogo (approved):
Super!

PR #307, opened 2021-09-10
Add Clang 13 for modern images

uilianries:
Clang 13 is still under development. Let's see if we need to update something


modern/.env
```diff
@@ -22,6 +22,7 @@ GCC11_VERSION=11.1.0
 CLANG10_VERSION=10.0.1
 CLANG11_VERSION=11.1.0
 CLANG12_VERSION=12.0.0
+CLANG13_VERSION=13.0.0-rc2
```
SSE4, line 25:
```suggestion
CLANG13_VERSION=13.0.0
```

Issue #304, opened 2021-08-26
[xenial] Configure Conan inside the image


jgsogo:
I'm proposing to run one extra step after installing Conan in the new images:

```
conan config init
conan profile update settings.compiler.libcxx=libstdc++11 default
``` 

IMHO, I think that `libstdc++11` is probably a better default if you just start to use one of these images and want to build something.

uilianries:
As it doesn't affect Conan Center, and libstdc++11 will be the default one for Conan 2.0, I agree.

PR #309, opened 2021-09-15
conan config: enable revisions, use libstdc++ by default in gcc

jgsogo:
IMO, it would be useful to use the defaults that are aligned with the future.

Closes https://github.com/conan-io/conan-docker-tools/issues/304

### DIFF
--- a/modern/clang/Dockerfile
+++ b/modern/clang/Dockerfile
@@ -155,4 +155,5 @@ RUN sudo mv /tmp/gcc/lib64 /usr/local/ \
     && sudo update-alternatives --install /usr/local/bin/ld ld /usr/local/bin/ld.lld 100 \
     && sudo rm /etc/ld.so.cache \
     && sudo ldconfig -C /etc/ld.so.cache \
-    && conan config init --force
+    && conan config init --force \
+    && conan config set general.revisions_enabled=1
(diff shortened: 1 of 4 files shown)


modern/clang/Dockerfile
```diff
@@ -155,4 +155,4 @@ RUN sudo mv /tmp/gcc/lib64 /usr/local/ \
     && sudo update-alternatives --install /usr/local/bin/ld ld /usr/local/bin/ld.lld 100 \
     && sudo rm /etc/ld.so.cache \
     && sudo ldconfig -C /etc/ld.so.cache \
-    && conan config init --force
```
uilianries, line 158: Why not libstdc++11 for clang too?

jgsogo: Or `libc++`? I have doubts, but for GCC I think there are no doubts, right?

uilianries: libstdc++11 I think will work fine, but libc++ has some limitations. We saw some projects which didn't build with libc++. Personally, I prefer libc++, but I would like to keep align/prepared for Conan 2.0 profiles.

jgsogo: Conan 2.0 is not going to autodetect anything, so that won't be a problem.

The issue I try to fix with GCC is the getting-started. I'm sure you have come across the ABI link error when using Conan many times... and you spend some time until you realize it is using _the wrong_ libcxx configuration. And it is wrong because the world already uses new ABI (the underlying system, binaries you can get from internet,...)

TBH, I don't use Clang, so I don't come across any error 🤷  But, I would say that `libc++` is more like the default for clang, so it would be expected if you start from scratch.

uilianries: > I'm sure you have come across the ABI link error when using Conan many times

True, that's why I use libstdc++11 by default now.

Let's update gcc for now and think about clang for a next clang.

uilianries (approved):

uilianries:
I won't consider that error for this PR. It's related to a exotic docker image for Linux + Mingw, plus, that error is related to Download.

PR #327, opened 2021-10-06
[ci] Improvements

jgsogo:
* Compute dynamically the Conan versions to generate for CCI
 * Ensure unique name is used for each execution of `_generate` pipeline.


uilianries:
I see there is a range of Conan versions, which should be from 1.37.2 (c3i_jenkins) to 1.40.4 (latest).

I only can see 1.40.3 (because 1.40.4 was released few hours ago), but where are older versions like 1.39.0?

PR #356, opened 2021-12-29
CI to generate docker images after a new Conan release

jgsogo:
So we don't forget anything whenever there is a new Conan release. This job will be triggered by the Conan release process.

### DIFF
--- a/.ci/on_conan_release.jenkinsfile
+++ b/.ci/on_conan_release.jenkinsfile
@@ -1,0 +1,87 @@
+/*
+
+    Triggered whenever a new Conan version is uploaded to pypi.
+
+*/
+
+node('Linux') {
+    stage('Input parameters') {
+        echo """
+        - conan_version: ${params.conan_version}
+        - build_new_images: ${params.build_new_images}
+        - build_old_images: ${params.build_old_images}
+        """
+    }
+
+    stage('New docker images') {
+        // Trigger the build (no need to wait)
+        if (params.build_new_images) {
+            build(job: 'ConanCenterIndex', wait: false)
+        }
+    }
+
+    stage('Old docker images') {
+        // Here we need to create a PR modifying the Conan version
+
+        // Clone, save and check if diff
+        checkout([$class           : 'GitSCM',
+                  branches         : [[name: '*/master']],
+                  userRemoteConfigs: [[credentialsId: 'conanci-gh-token', url: 'https://github.com/conan-io/conan-docker-tools.git']],
+                  extensions       : [[$class           : 'RelativeTargetDirectory',
+                                       relativeTargetDir: 'conan-docker-tools']],
+        ])
+
+        // Modify files that list Conan version. Change defaults to match this latest version
+        def content1 = readFile(file: 'conan-docker-tools/build.py')
+        content1 = content1.replaceAll(/TARGET_CONAN_VERSION = "\d+\.\d+.\d+"/, "TARGET_CONAN_VERSION = \"${params.conan_version}\"")
+        writeFile(file: 'conan-docker-tools/build.py', text: content1)
+
+        def content2 = readFile(file: 'conan-docker-tools/modern/.env')
+        content2 = content2.replaceAll(/CONAN_VERSION=\d+\.\d+.\d+/, "CONAN_VERSION=${params.conan_version}")
+        content2 = content2.replaceAll(/DOCKER_TAG=\d+\.\d+.\d+/, "DOCKER_TAG=${params.conan_version}")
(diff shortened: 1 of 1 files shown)


.ci/on_conan_release.jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+        String branchName = "bot/release-${params.conan_version}".replaceAll(/\./, '_')
+        dir('conan-docker-tools') {
+            sh "git checkout -b ${branchName}"
+            sh "git diff"
+
+            sh 'git config --local user.email "javierg@jfrog.com"'
+            sh 'git config --local user.name "Release Bot"'
+            sh 'git commit -am "[up conan] Update Conan version"'
+
+            if (params.build_old_images) {
```
danimtb, line 54: This condition should be checked earlier and skip the whole stage if not true

.ci/on_conan_release.jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+        - conan_version: ${params.conan_version}
+        - build_new_images: ${params.build_new_images}
+        - build_old_images: ${params.build_old_images}
+        """
+    }
+
+    stage('New docker images') {
+        // Trigger the build (no need to wait)
+        if (params.build_new_images) {
+            build(job: 'ConanCenterIndex', wait: false)
```
danimtb, line 19: Don't you have to pass the conan_version to the job or the latest version is picked automatically?

uilianries: I believe it will be filled by the Conan Release Jenkins job.

.ci/on_conan_release.jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+
+    Triggered whenever a new Conan version is uploaded to pypi.
+
+*/
+
+node('Linux') {
+    stage('Input parameters') {
+        echo """
+        - conan_version: ${params.conan_version}
+        - build_new_images: ${params.build_new_images}
```
uilianries, line 11: Could you comment in this code, how these 3 variables are filled?

.ci/on_conan_release.jenkinsfile
jgsogo: I use it to the pipeline, this way I can check if all the process works and I just skip the actual action 😅

.ci/on_conan_release.jenkinsfile
jgsogo: The `ConanCenterIndex` one takes the latest dockerfiles from the repository and generates images for the latest Conan versions from pypi. That way we ensure that all of them are identical (except for the Conan version).

So here there is no need for that version.

.ci/on_conan_release.jenkinsfile
jgsogo: These are parameters/inputs to the Jenkins job. We are going to trigger it Jenkins shows the appropriate form: 
image

If called from other jobs, they just need to provide the values.

.ci/on_conan_release.jenkinsfile
```diff
@@ -0,0 +1,87 @@
+/*
+
+    Triggered whenever a new Conan version is uploaded to pypi.
+
+*/
+
+node('Linux') {
+    stage('Input parameters') {
```
uilianries, line 8:
```suggestion
    stage('Input parameters') {
    // These are parameters/inputs to the Jenkins job. They are provided by onConanVersionReleased.
    // If called from other jobs, they just need to provide the values.
```

danimtb (approved):

.ci/on_conan_release.jenkinsfile
```diff
@@ -0,0 +1,87 @@
...
+        writeFile(file: 'conan-docker-tools/build.py', text: content1)
+
+        def content2 = readFile(file: 'conan-docker-tools/modern/.env')
+        content2 = content2.replaceAll(/CONAN_VERSION=\d+\.\d+.\d+/, "CONAN_VERSION=${params.conan_version}")
+        content2 = content2.replaceAll(/DOCKER_TAG=\d+\.\d+.\d+/, "DOCKER_TAG=${params.conan_version}")
+        writeFile(file: 'conan-docker-tools/modern/.env', text: content2)
+
+
+        String branchName = "bot/release-${params.conan_version}".replaceAll(/\./, '_')
+        dir('conan-docker-tools') {
```
uilianries, line 46:
```suggestion
        // TODO: Check if this is already created, in case the stage `Open PR in Github` fails.
        dir('conan-docker-tools') {
```

uilianries (approved):

PR #358, opened 2022-01-13
Update git version

### LINKED ISSUE #357: [rfc] Upgrade Git version
I´ve been using some of the new images with Github actions, and realized they use quite old Git version `2.8` (the one for Ubuntu 16.04 in new images). Some actions (`actions/checkout@v2`) requires newer Git version.

I think nothing is stopping us from using _latest_ Git in all our docker images. There is no restriction on Conan client-side AFAIK.

uilianries:
I didn't update all images, because we are rotating new Docker images on Conan Center, otherwise, it would consume much more time and I don't there are many users running old images.

UPDATE: Fixing package version on Ubuntu doesn't, when a new package version is available for update, the older version is removed, so or we install the only available version or any build can break.

closes #357

/cc @Croydon 


jgsogo (changes requested):
At least we need to update all the images we are currently using in CC (old docker images):
 * GCC: "5", "7", "8", "9", "10"

uilianries:
@jgsogo I'll update it

modern/tests/test_installed/test_installed.py
```diff
@@ -9,7 +9,7 @@
     "perl": {"16.04": "5.22.1"},
     "wget": {"16.04": "1.17.1"},
     "curl": {"16.04": "7.47.0"},
-    "git": {"16.04": "2.7.4"},
+    "git": {"16.04": "2.34.1"},
```
jgsogo, line 12: Probably new versions will arrive to `ppa:git-core/ppa`, so we need to install a fixed one.

Can we use in the docker command something like...
```
apt-get install -y --no-install-recommends --no-install-suggests \
   ...
   git=2.34.1
   ...
```

uilianries: Not needed. There is not option when a version become available on PPA. It's like Highlader, only one can exist. https://launchpad.net/~git-core/+archive/ubuntu/ppa?field.series_filter=xenial When a new version become available, 2.34.1 will be removed from there and will break the installation. It's a normal process.  The only way to keep more than one version, is uploading deb packages to Artfactory, but it's something that not inclined to manage.

jgsogo: Then we should change the test for something like `>=2.34.1`. The CI has to work. Being `git` I'm pretty sure it will keep working, so probably there is no need to run Git commands... but just in case I would also check major version `<3`

jgsogo:
We should also generate the (old) docker images after this PR, so they can also benefit from this new Git version

uilianries:
> We should also generate the (old) docker images after this PR, so they can also benefit from this new Git version

I didn't get. Jenkins does not generate it after merging?

jgsogo:
No. For new docker images we can connect it (right now we need to click the button, is not automatic).

For old docker images the CI uses the Conan version that is committed to the repository, so we should commit and push (and wait) for each one of the version we want to generate, right?

uilianries:
So we need a diff based on Conan version to trigger a new build for older images? It sounds bad, if we need a hotfix, it will be hard to be done.

jgsogo:
**We need consistency** across all the images that we are using in CCI. If one of them uses git=2.34.1, all of them should use that version too. If now we are using those belonging to Conan v1.43.1, they should start to use git=2.34.1 (there is a reason why we decided to change that version, or merge any other PR). And, if tomorrow, we start to use v1.43.3, we expect they use git=2.34.1... and two weeks later when we move to use v1.44.1 we want them to run git=2.34.1.

This is only possible if, whenever we merge something to `master`, we generate the images for all the Conan version that we are already using or we might use in the future. That way we are somehow decoupling this repository from Conan releases and from CCI, each one can move at its own pace.

PR #370, opened 2022-02-02
Update binutils for GCC11

uilianries:
closes #367

We also can build from sources, but there is an error when installing:

    libctf.so.0: version `LIBCTF_1.1' not found (required by /usr/bin/ld)

```
RUN wget --no-check-certificate --quiet -O /tmp/binutils-2.36.tar.xz https://ftp.gnu.org/gnu/binutils/binutils-2.36.tar.xz \
    && tar Jxf /tmp/binutils-2.36.tar.xz -C /tmp \
    && cd /tmp/binutils-2.36 \
    && sed -i '/@\tincremental_copy/d' gold/testsuite/Makefile.in \
    && mkdir /tmp/binutils-2.36/build \
    && cd /tmp/binutils-2.36/build/ \
    && ../configure --prefix=/usr --enable-gold --enable-ld=default --enable-plugins --enable-shared --disable-werror --enable-64-bit-bfd --with-system-zlib \
    && make -s tooldir=/usr --jobs=$(nproc) \
    && make -s tooldir=/usr install
```

/cc @SSE4 @jgsogo 

- [ ] Refer to the issue that supports this Pull Request.
- [ ] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [ ] I've read the [Contributing guide](https://github.com/conan-io/conan-docker-tools/blob/master/.github/CONTRIBUTING.md).
- [ ] I've followed the [PEP8](https://www.python.org/dev/peps/pep-0008) style guides for Python code.
- [ ] I've followed the [Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) guides for Dockerfile.

PR #372, opened 2022-02-02
Git - test minimum version

jgsogo:
After https://github.com/conan-io/conan-docker-tools/pull/358 it was expected that the new Git version would break current tests.

Here I'm modifying the tests, now we check the minimum boundary for Git, the upper boundary is open.

---

NOTE.- If anyone knows how to install a fixed version, it would be great to hear it 🙌

### DIFF
--- a/modern/tests/fixtures/expected.py
+++ b/modern/tests/fixtures/expected.py
@@ -1,45 +1,10 @@
 import os.path
 import re
+import pytest
 from collections import defaultdict
 from dataclasses import dataclass
 
-import pytest
-
-
-@dataclass
-class Version:
-    full_version: str
-    major: str
-    minor: str = None
-    patch: str = None
-
-    def __init__(self, full_version=None):
-        if not full_version:
-            return
-
-        self.full_version = full_version
-        self.major, *rest = self.full_version.split('.', 1)
-        if rest:
-            self.minor, *rest = rest[0].split('.', 1)
-            if rest:
-                self.patch = rest[0]
-
-    def __str__(self):
-        ret = f"{self.major}"
-        if self.minor is not None:
-            ret += f".{self.minor}"
-        if self.patch is not None:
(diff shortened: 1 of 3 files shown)


uilianries:
To install a fixed version you have to pass ==, like `apt install git==<version>` But again, it doesn't work on Ubuntu, if they update the version, they also remove the current version, then your installation is broken. We had a series of problems in the past trying to use fixed versions, at some point, Luis and me agreed in don't using fixed versions.

jgsogo:
We need fixed versions as much as possible. Everything around `conan-center-index` is based on stability and reproducibility. If we introduce here some change and things start to fail in that CI we will have a problem hard to find (no one changed anything) and hard to fix (how to downgrade the version to the one that works). That issue could lead to several days of downtime in CCI until we are able to find, implement and deploy the fix. We don't want that to happen 😄

uilianries:
Again, we don't have that control. Once ubuntu updates a package version, it will need to update or we will be broken. You can enforce a version, but it won't help as soon we have a new update, you will need to update and that's it. The only advantage is keeping a git history about which version was used during that time, but is something we can acquire form ubuntu packages too.
If you want to reproduce everything and lock versions, my suggestion is building from source, distros are not so cool like Conan, they only support the latest version.

jgsogo:
That's why I left the comment in the description, in case anyone knows how to do it with the system package manager. 

When we started to develop these modern images we agreed that we were not going to use Conan to install binaries (because of the chicken-and-egg problem), but after https://github.com/conan-io/conan-docker-tools/pull/370, maybe we decide to use it and we change our minds. If that's the case, we could install Git using Conan, and other tools as well.

jgsogo:
After merging (and CI passes in `master`), we need to trigger the generation of these docker images for ConanCenter.

uilianries (commented):
LGTM

PR #375, opened 2022-02-10
Install binutils from GCC5

uilianries:
Related to https://github.com/conan-io/conan-docker-tools/pull/370/files#r803628721

- Why profile file? 
  - To generate that lockfile we run:

    conan lock create -r conancenter -pr profile --lockfile-out=conan.lock conanfile.txt

### DIFF
--- a/modern/base/conan.lock
+++ b/modern/base/conan.lock
@@ -48,5 +48,5 @@
   "revisions_enabled": true
  },
  "version": "0.4",
- "profile_host": "[settings]\narch=x86_64\narch_build=x86_64\nbuild_type=Release\ncompiler=gcc\ncompiler.libcxx=libstdc++\ncompiler.version=5\nos=Linux\nos_build=Linux\n[options]\n[build_requires]\n[env]\n"
+ "profile_host": "[settings]\narch=x86_64\narch_build=x86_64\nbuild_type=Release\ncompiler=gcc\ncompiler.libcxx=libstdc++11\ncompiler.version=5\nos=Linux\nos_build=Linux\n[options]\n[build_requires]\n[env]\n"
 }
--- a/modern/base/conan.lock
+++ b/modern/base/conan.lock
@@ -23,15 +23,15 @@
    "2": {
     "ref": "zlib/1.2.11#b46bf663179824c23085bb71666ba8e6",
     "options": "fPIC=True\nshared=False",
-    "package_id": "dfbe50feef7f3c6223a476cd5aeadb687084a646",
-    "prev": "f064738ed9ea565f75cd7ebc20c2be62",
+    "package_id": "76f87539fc90ff313e0b3182641a9bb558a717d2",
+    "prev": "a4730ad7fc025658ed1806f0b0d5ba1f",
     "context": "host"
    },
    "3": {
-    "ref": "readline/8.0#1d21941e543bef7a06551f5a56c8a87e",
+    "ref": "readline/8.0#6bfe51715475f7b96c43adfa7b6aef96",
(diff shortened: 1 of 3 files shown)


modern/base/profile
```diff
@@ -0,0 +1,13 @@
+[settings]
+os=Linux
+os_build=Linux
+arch=x86_64
+arch_build=x86_64
+compiler=gcc
+compiler.version=5
+compiler.libcxx=libstdc++
```
jgsogo, line 8: old ABI?

uilianries: let me update to libstdc++11

uilianries: Done

jgsogo (approved):
IMHO, better to use GCC5 binaries, are they are linked using the lowest stdlibc++ version... just in case we add some C++ application in the future.

wdyt, @SSE4 ? is my understanding right?

SSE4:
I think it should be irrelevant, as bintutils are pure C (at least, ld and gas) AFAIK and shouldn't introduce a link to libstdc++.

```
conan@08023348d030:~$ ldd $(which as)
	linux-vdso.so.1 =>  (0x00007ffdb5daa000)
	libopcodes-2.26.1-system.so => /usr/lib/x86_64-linux-gnu/libopcodes-2.26.1-system.so (0x00007f41bc865000)
	libbfd-2.26.1-system.so => /usr/lib/x86_64-linux-gnu/libbfd-2.26.1-system.so (0x00007f41bc520000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f41bc306000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f41bbf3c000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f41bbd38000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f41bcc0b000)
conan@08023348d030:~$ ldd $(which ld)
	linux-vdso.so.1 =>  (0x00007ffc165e7000)
	libbfd-2.26.1-system.so => /usr/lib/x86_64-linux-gnu/libbfd-2.26.1-system.so (0x00007f302803d000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f3027e39000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3027a6f000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f3027855000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3028382000)
```
but for some foreseen future it should be fine

modern/base/conan.lock
```diff
@@ -40,13 +40,13 @@
    "4": {
     "ref": "termcap/1.3.1#a75d9f61bf0d1899b64967958dfa6538",
     "options": "fPIC=True\nshared=False",
-    "package_id": "dfbe50feef7f3c6223a476cd5aeadb687084a646",
-    "prev": "383d0c76a636c07aa78d0065f1e0a6f8",
+    "package_id": "76f87539fc90ff313e0b3182641a9bb558a717d2",
+    "prev": "34552acd6624780a5f6251750f40d9c5",
```
SSE4, line 44: why package ID has changed?

uilianries: Compiler version, from GCC11 to GCC5.

SSE4: okay, so it's not removed in `package_id`...

SSE4 (approved):

PR #393, opened 2022-03-24
Install python 3.7.13 for all Docker images

### LINKED ISSUE #392: upgrade CCI images to Python 3.7

python 3.6 is EOL. we need to upgrade to 3.7. @uilianries 

### Environment Details
* Conan Docker Tools image: **conanio/gcc7:1.13.1**
* Operating System: **Linux**
* Operation System Version: **Ubuntu 16.04**
* Python version (If you are building some image): **python 3.7.0**

uilianries:
Latest Python 3.7 patch available is 13: https://www.python.org/downloads/release/python-3713/

closes #392 

we need at least Python 3.7 to use new versions of Meson and we need Meson in CCI to build some packages.


Croydon (approved):

SSE4 (approved):

uilianries:
Need to get aligned with Windows and Mac machines before pushing.

It will need to generate new docker images updated.

jgsogo:
Strictly there is no need to wait. Docker images (legacy ones) were a party with so many different Python versions, moving all of them to the same version is good, and eventually, Windows and Mac will be moved from 3.6 to 3.7. This change shouldn't break anything (given that the recipes were already using 3.6 for some configurations and 3.7 should be valid for all).

Anyway, please, state the reason for this change in the PR: it is not because EOL of Python 3.6, it is because we need at least Python 3.7 to use new versions of Meson and we need Meson in CCI to build some packages.

jgsogo (approved):

PR #396, opened 2022-04-05
Remove 32-bit support from Clang 11 legacy image

uilianries:
- It seems be Ubuntu dropped i386 apt folder
- We don't use libstdc++ multiarch with Clang 11 docker image


legacy/clang_11/Dockerfile
```diff
@@ -55,6 +40,14 @@ RUN dpkg --add-architecture i386 \
        libncurses5-dev \
        libncursesw5-dev \
        liblzma-dev \
+    && wget https://apt.llvm.org/llvm.sh \
```
SSE4, line 43: that's super unsecure to run script downloaded from the intertet
at least checking sha of the file would be good

uilianries: It comes directly from llvm, but still, I had a certificate error on Mac. I'll collect what it's doing and push to the Dockerfile directly (it's updating apt files only). wget llvm was not a good idea, even as recommended the by official website here https://apt.llvm.org/

uilianries: Done, now it uses the regular approach

PR #435, opened 2022-05-26
Patch lsb_release for Clang 10 and Clang 11

uilianries:
Related to https://github.com/conan-io/conan-center-index/pull/10795

lsb_release is an APT package which wants the python from the system, not from pyenv, which causes a mess when trying to execute from pyenv version.

As workaround, we link the original script to pyenv folder, so we "install" lsb_release to our global python version.

We already did it in the past, but forgot to add for new clang images. For modern images it's not required, lsb_release works there.


SSE4 (approved):

maksim-0:
@uilianries Could you please clarify when new images will be available in c3i?

uilianries:
@maksim-0 they should be available now. 1.48.0 is still building, but 1.47.0 was uploaded minutes ago

PR #438, opened 2022-06-02
[up conan] Update Conan version (1.48.2)

conanci:
New Conan version released.

---

<sup>This PR has been created automatically by CI</sup>


jgsogo (changes requested):
The PR is created for every new Conan release, but we should keep here the latest one (human inspection).

PR #444, opened 2022-06-08
Remove system dev packages for modern images

uilianries:
All dev ubuntu packages should be consumed from Conan Center, otherwise, our package can consume from system by mistake.

All removed packages are present in Conan Center.

Let's start with modern, then, we can go with legacy.

/cc @jcar87

### DIFF
--- a/modern/base/Dockerfile
+++ b/modern/base/Dockerfile
@@ -93,7 +93,21 @@ RUN apt-get -qq update \
     && mv /etc/ld.so.conf.d/libc.conf /etc/ld.so.conf.d/10libc.conf \
     && rm /etc/ld.so.cache \
     && ldconfig -C /etc/ld.so.cache \
-    && apt-get -qq purge -y g++-multilib gcc \
+    && apt-get -qq purge -y \
+       g++-multilib \
+       gcc \
+       libgmp-dev \
+       libmpfr-dev \
+       libmpc-dev \
+       libffi-dev \
+       libssl-dev \
+       zlib1g-dev \
+       libbz2-dev \
+       libsqlite3-dev \
+       libreadline-dev \
+       libncurses5-dev \
+       libncursesw5-dev \
+       liblzma-dev \
     && apt-get -qq autoremove -y \
     && apt-get -qq autoclean \
     && apt-get -qq update \


modern/base/Dockerfile
```diff
@@ -93,7 +93,21 @@ RUN apt-get -qq update \
...
     && ldconfig -C /etc/ld.so.cache \
-    && apt-get -qq purge -y g++-multilib gcc \
+    && apt-get -qq purge -y \
+       g++-multilib \
+       gcc \
+       libgmp-dev \
+       libmpfr-dev \
+       libmpc-dev \
+       libffi-dev \
+       libssl-dev \
```
Croydon, line 103: Great idea, but aren't some of them, like libssl dependencies that are also used by Python? 🤔

uilianries: That's a good question which I should validate before merging it. Not sure if we built python using static libraries or shared, but I gonna check it. Thank you for reviewing it!

jcar87: Indeed, it's a good idea to validate and make sure.

In typical Debian convention, the runtime and files required for development are split in separate packages. The packages being removed are the ones with the `-dev` suffix - these include the header files, static libraries, and sometimes the "unversioned" `.so` file that is a symlink to the one in included in the runtime package.

So this shouldn't affect the _runtime_ libraries that may be required by python and so on.

However, some `pip install` commands may cause some binaries to be built from source, using the default system paths, so they may expect some headers to live in the system - this is something worth checking!

jcar87: In recent years (I'd say the past 2-3 years) `pip` has been improving a lot with regards to pre-packaged binaries inside wheels. However for a given version of the interpreter and a variant of Linux/architecture, the availability of certain packages in binary form may be patchy. Some packages are not available in PyPI in binary form and are always built from source, but my experience so far is that things have been improving.

We could try and pass the `--prefer-binary` option during `pip install` (docs [here](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-prefer-binary)) - the resolver should still pick up a compatible version, but it would give priority to versions that have pre-packaged binaries rather than those that need building from source.  

A perhaps less-elegant solution would be to wrap the `RUN pip install` commands with `apt-get install` and `apt-get remove` before and after invoking pip install, in a way that potential system dependencies are available, but they `-dev` packages are still removed and don't make it into the images.

uilianries: All good! 

I tested with some packages with require those system packages, which are using conan instead:
- oatpp-sqlite
- readline
- cpython
- bliss
- mpfr

No errors, built with shared and static.

PR #446, opened 2022-06-13
[legacy] Remove Ubuntu dev packages

uilianries:
Update GCC and Clang legacy images, removing the -dev from ubuntu reporistory


jcar87 (approved):
Brilliant!

PR #461, opened 2022-08-31
[hotfix] GCC build option: nsl should be nls instead

### LINKED ISSUE #460: Wrong option spelling when building gcc?

When building GCC `disable-nsl` option is used:
https://github.com/conan-io/conan-docker-tools/blob/4d3730dde48a819e531125a54d99f67f91663ecc/modern/gcc/Dockerfile#L39

I don't see it there https://gcc.gnu.org/install/configure.html. Shouldn't it be `disable-nls` instead?

uilianries:
Fix: Wrong option disabled for GCC build is not check by configure. Maybe GCC bootstrap does it, but it would increase our build time by 1h at least.

closes #460

/cc @maksim-petukhov


jgsogo (approved):
Looks like a typo. Thanks!

PR #474, opened 2022-10-20
support arm build

boussaffawalid:
Changelog: (Feature | Fix | Bugfix): Describe here your pull request

### DIFF
--- a/modern/base/Dockerfile
+++ b/modern/base/Dockerfile
@@ -29,7 +29,7 @@ RUN apt-get -qq update \
        binutils \
        wget \
        git \
-       g++-multilib \
+       g++ \
        libgmp-dev \
        libmpfr-dev \
        libmpc-dev \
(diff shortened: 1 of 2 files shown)


uilianries:
@boussaffawalid Thank you for your interesting on supporting ARM, but as documented on [modern images readme](https://github.com/conan-io/conan-docker-tools/tree/master/modern#conan-docker-tools), those images are exclusive for Conan Center Index usage, so we will not add extra community support, including new archs or OS (Android) because is not planned for now.

You still can fork and publish your Docker images on hub.docker.io, or, you can consume legacy GCC10 images prepared for Arm. They are no longer supported, but I updated them one month ago, with a second profile for cross-building.

- https://github.com/conan-io/conan-docker-tools/tree/master/legacy/gcc_10-armv8
- https://github.com/conan-io/conan-docker-tools/tree/master/legacy/gcc_10-armv7
- https://github.com/conan-io/conan-docker-tools/tree/master/legacy/gcc_10-armv7hf

Said that, I'm closing this PR as supporting extra configuration is not planned in a short-term. Thank you!

boussaffawalid:
@uilianries we use  both `x86_64` and `arm` so instead of having `x86_64-linux-gnu` hardcoded in the dockerfile, I used the output of `dpkg-architecture --query DEB_BUILD_GNU_TYPE` which will return the correct architecture. 

Just keeping the dockerfile generic  in case somebody else needed this :)

PR #505, opened 2023-04-30
Update GCC Compiler Minor Versions

### LINKED ISSUE #504: Proposal to Update GCC Minor Versions
Proposal to Update GCC Minor Versions to the Latest Available:
- [ ]  8.4.0 -> 8.5.0
- [ ] 9.3.0 -> 9.5.0
- [ ] 10.3.0 -> 10.4.0
- [ ] 11.1.0 -> 11.3.0
- [ ] 12.1.0 -> 12.2.0

KaseyJenkins:
Changelog: (Feature):

"fixes #504"

- [x] 8.4.0 -> 8.5.0
- [x] 9.3.0 -> 9.5.0
- [x] 10.3.0 -> 10.4.0
- [x] 11.1.0 -> 11.3.0
- [x] 12.1.0 -> 12.2.0


uilianries:
@KaseyJenkins Thank you for your contribution. Updating docker is really critical as affects all packages stored in Conan Center. So I need to ask if you have any blocker in Conan Center Index, that is not possible to build a recipe, for instance. Otherwise, I would prefer being more conservative and not updating the minor version now.

uilianries:
@KaseyJenkins It would be nice having gcc 13 yes. We just didn't include yet because is not asked to be used on CCI (yet), because would need to build all packages there and requires days of CI to build everything. 

Plus, clang is also missing versions 15 and 16.

However, only the modern folder should be updated. The legacy images should no longer be updated. Thank you for asking!

PR #506, opened 2023-05-11
[clang11] Use Ubuntu focal (20.04) as base image

uilianries:
The image Bionic (18.04) provides libstdc++-6.so on the version 3.4.25 which does not allow using C++17 libraries like `std::filesystem`. Unfortunately, it's not possible to obtain a newer version directly from Ubuntu Bionic official repository.

As solution, we can update the base distro to Focal (20.04) that contains a newer version of libstd++ 3.4.28 and supports C++17.

The same for glibc version, Bionic supports 2.27 and Focal provides 2.31.

As both glibc and libstdc++ are backward compatible, it should not break CCI.

Doing a local test, by consuming packages produced by the current Docker image `conanio/clang11:1.59.0` in a new image based on Focal, it worked without errors.

Related to https://github.com/conan-io/conan-center-index/pull/16938

/cc @czoido 

### DIFF
--- a/legacy/clang_11/Dockerfile
+++ b/legacy/clang_11/Dockerfile
@@ -40,7 +40,7 @@ RUN apt-get -qq update \
        libncurses5-dev \
        libncursesw5-dev \
        liblzma-dev \
-    && printf 'deb http://apt.llvm.org/bionic/   llvm-toolchain-bionic-11  main' >> /etc/apt/sources.list \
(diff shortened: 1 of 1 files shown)


Croydon:
Even the `modern` images use only 18.04 right now. And why is CCI not switching to those instead of updating the old images?

uilianries:
We opted by dropping Clang 11 instead of keeping legacy images working: https://github.com/conan-io/conan-center-index/pull/17527

/cc @Croydon

PR #509, opened 2023-05-19
Remove older compiler versions from build list

uilianries:
Some compilers are no longer used by CCI and may cause some errors when building their docker images, only resulting in more trouble. This PR reduces the number of docker images produced by the CI, to avoid those failures.  

GCC 6 is no longer supported for any docker image.

Only Clang 11 will be supported for legacy images. 


legacy/.env
```diff
@@ -1,8 +1,8 @@
-CONAN_VERSION=2.0.5
-DOCKER_BUILD_TAG=2.0.5
+CONAN_VERSION=1.60.0
```
uilianries, line 1: Need to open a second PR, after merging this one, to build 2.0.5

RubenRBS (approved):

PR #531, opened 2023-07-25
gcc: always pass --build-id to linker

marxin:
As mentioned in
https://github.com/conan-io/conan-center-index/issues/18809, currently some conan shared libraries are build without .note.gnu.build-id note and thus one cannot link separated debug info ELF file with the shared library. It is something CPack complains about: https://github.com/Kitware/CMake/blob/master/Modules/Internal/CPack/CPackDeb.cmake#L231

GCC documentation of the added option:
https://gcc.gnu.org/install/configure.html

Changelog: Bugfix: Describe here your pull request

- [ ] Fixes https://github.com/conan-io/conan-center-index/issues/18809 where some Conan packages with shared libraries are missing `.note.gnu.build-id ` section

### DIFF
--- a/modern/gcc/Dockerfile
+++ b/modern/gcc/Dockerfile
@@ -49,6 +49,7 @@ RUN cd gcc-${GCC_VERSION} \
                    --without-isl \
                    --with-system-zlib \
                    --prefix=/tmp/install \
+                   --enable-linker-build-id \
     && make -s -j$(nproc) \
     && make install-strip
 


marxin:
Note clang has a similar build option `-DENABLE_LINKER_BUILD_ID` which should be enabled as well.

uilianries (approved):
Okay, found both Ubuntu and Arch using this flag:

- https://git.launchpad.net/ubuntu/+source/gcc-10/tree/debian/rules2?h=ubuntu/focal-devel#n226
- https://gitlab.archlinux.org/archlinux/packaging/packages/gcc/-/blob/main/PKGBUILD#L92

uilianries:
@marxin Thank you for your contribution. The CI should build news images and will be available in some hours.

marxin:
> About Clang, do you want me to create another pull request?

Done as #532.

uilianries:
> Do you know if the affected package will be re-built once the new images are created? Or should I request it with an issue?

No. Packages are only built with a new PR or, by direct action as an internal CI build. I can build it again internally, but what package do you need, is `libiconv/1.17` ?

marxin:
Well, right now, I am affected by the following shared libs:
```
ls *.so* | xargs file | grep ELF | grep -v BuildID                                                                                                                                                                                                                                                     
libcharset.so.1.0.0:                        ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
libgeos.so.3.11.0:                          ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, not stripped
libgeos_c.so.1.17.0:                        ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, not stripped
libiconv.so.2.6.1:                          ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
liblaszip.so.8.0.5:                         ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
liblzma.so.5.2.5:                           ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
libopenjp2.so.2.5.0:                        ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
libshp.so.1.5.0:                            ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
libtbb.so.2:                                ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
libtbbmalloc.so.2:                          ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
libtbbmalloc_proxy.so.2:                    ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

but once I update my dependencies, I can be affected by different packages (or versions).
What about connecting the Build-Id problem with #505 where a mass rebuild can be triggered in the future (that will fix both missing Build-Id for Clang and GCC)?

uilianries:
> where a mass rebuild can be triggered in the future (that will fix both missing Build-Id for Clang and GCC)?

I would prefer keeping rebuilding tasks in Conan-Center-Index repository, so more people will know about it. About doing a mass rebuild is possible, but need to study it first, as the build order is important here. I'll take a look into this on next few days, right now CCI service is broken 😞 Thank you for sharing the list.

marxin:
Can you please trigger rebuild for the following packages that still miss the `BuildID`:

```
gklib/5.1.1
easyexif/1.0
libffi/3.4.4
giflib/5.2.1
gtest/1.10.0
sqlite3/3.37.1
xerces-c/3.2.3
```

PR #536, opened 2023-08-03
[ci] Accept version list by parameter

uilianries:
Enable the feature of passing a specific list of Conan client versions to be built.


uilianries:
Already implemented.

PR #554, opened 2023-12-05
Add the packaging and mako Python modules

jwillikers:
Changelog: Feature:

The packaging Python module is required for glib.
This dependency was added [here](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3740) in order to support Python 3.12 which removed the distutils module.
conan-io/conan-center-index#21535 is a backport of this change for CCI which allows building recent versions of glib with Python 3.12.

The mako Python module is required for the Mesa Conan package in conan-io/conan-center-index#20528.

I'm not sure if I need to update the legacy images as well.
Please let me know if I should.

### DIFF
--- a/modern/base/Dockerfile
+++ b/modern/base/Dockerfile
@@ -124,4 +124,4 @@ WORKDIR /home/conan
 RUN pip install conan~=1.0 \
     && CONAN_REVISIONS_ENABLED=1 CONAN_USER_HOME=/tmp/conan conan install -r conancenter -g deploy -if /opt/conan -l /opt/conan/conan.lock /opt/conan/conanfile.txt \
     && rm -rf /tmp/conan \
-    && pip install conan==${CONAN_VERSION} packaging==23.2
+    && pip install conan==${CONAN_VERSION} mako==1.3.0 packaging==23.2
--- a/modern/base/Dockerfile
+++ b/modern/base/Dockerfile
@@ -124,4 +124,4 @@ WORKDIR /home/conan
 RUN pip install conan~=1.0 \
     && CONAN_REVISIONS_ENABLED=1 CONAN_USER_HOME=/tmp/conan conan install -r conancenter -g deploy -if /opt/conan -l /opt/conan/conan.lock /opt/conan/conanfile.txt \
     && rm -rf /tmp/conan \
-    && pip install conan==${CONAN_VERSION}
+    && pip install conan==${CONAN_VERSION} packaging==23.2


uilianries:
I don't get, we are running Python 3.6, not 3.12. 🤔

jwillikers:
> I don't get, we are running Python 3.6, not 3.12. 🤔

The `packaging` module will be required to build glib regardless of the Python version. The change to require this module was due to a change in Python 3.12.

ericLemanissier:
@uilianries @danimtb @RubenRBS What do you think of this change ?
The motivation is simple: glib >= 2.79.0 requires `packaging` https://gitlab.gnome.org/GNOME/glib/-/commit/e597b189c36651d83dd72dfdc8530682c80fc235#9f621eb5fd3bcb2fa5c7bd228c9b1ad42edc46c8_1_204
Mesa requires mako https://github.com/Mesa3D/mesa/blob/31c9e17bf2a5e3ff73fce7a2cf14db121c5bbc90/docs/install.rst#third-partyextra-tools

If you agree, this needs to be installed on macos and windows runners too

RubenRBS:
Prioritizing this on monday, thanks for the heads-up :)

ericLemanissier:
Note that this PR is a solution only for C3I.
Another solution is to create a system package which installs python3-packaging, so that it also fixes consumers using the system python. It still would not help consumers using a non-system provided python.
We could also consider a system package calling `pip install packaging` but pip is not the only provider for Python packages, and it may not even be installed...


Regarding the usefulness to have packaging and mako, packaging is close to mandatory because it replaces a python2 deprecated module: distutils. It has a lot of dependants: https://ubuntu.pkgs.org/23.10/ubuntu-main-amd64/python3-packaging_23.1-1_all.deb.html
Mako is less important, but unfortunately it's been a requirement of Mesa for as long as I remember.  It has fewer dependants https://ubuntu.pkgs.org/23.10/ubuntu-main-amd64/python3-mako_1.2.4+ds-2_all.deb.html

uilianries:
@ericLemanissier Yes, good point. We also don't want to install in the CI, because there is the case where people can download Conan client as a standalone executable. I sent an issue to the upstreams, we could install from meson.build, but not sure if they will accept. Another option is providing a Conan package to them, not good too, but not the worst.

ericLemanissier:
note that for conan client as standalone, there is no technical possibility to add any python package that was not already included in the standalone distribution. If you want to be able to build glib>=2.79.0 or mesa with the standalone executable, either conan must add packaging and mesa to its distribution, or conan-center should provide a package providing python and packages providing packaging and mako (feasability has to be proven)

memsharded:
I don't think that Conan pip package or self-contained bundle should contain some Python dependencies, because one build system requires them to package some specific libraries.

ericLemanissier:
actually, it may be a lot easier than I originally considered :
- https://github.com/conan-io/conan-center-index/pull/23091
- https://github.com/conan-io/conan-center-index/pull/23093

jwillikers:
I've opened the PR conan-io/conan-center-index#23098 to add Mako.

ericLemanissier:
Taking a look at the outside world:

vcpkg is planning to call `pip install packaging` during the build of glib:
- https://github.com/microsoft/vcpkg/pull/37473/files#diff-529f8d22a7b04e00f58b64a9bf1faad68595b34644aa699dc23408b51720e34aR42
- https://github.com/microsoft/vcpkg/blob/8d3649ba34aab36914ddd897958599aa0a91b08e/ports/vcpkg-get-python-packages/x_vcpkg_get_python_packages.cmake#L110

Conda package for glib requires `packaging` package, which also does `pip install`:
- https://github.com/conda-forge/glib-feedstock/blob/08f2436c08a9e83735ba5ea96650bf17530eaaf1/recipe/meta.yaml#L46
- https://github.com/conda-forge/packaging-feedstock/blob/8254020c8425d6e4b7566c6f1fac0a08f547f895/recipe/meta.yaml#L14

uilianries:
@jwillikers @ericLemanissier Thank you both for not only bringing this topic, but also investigating possible solutions. 

After taking to Conan team more than once to discuss this topic, and investigating, both projects, the usage of libraries, the possible inclusion and solutions in ConanCenterIndex, and what other package managers and build tools are doing, we concluded the follow decision:

- Installing these packages in Docker images is **not** an option. The docker images are for general propose and agnostic to any package in ConanCenterIndex. If we allow it today, so anyone can ask for other system packages later (In very past we did for Conan 1.x and resulted in hell dependencies)
- Creating Conan packages for python-mako and python-packaing is **not** option. These are very particular cases, and providing packages in CCI in complicated for some reasons:
  - Once merged, an window will be opened in ConanCenterIndex to accept more python package as Conan package, something that Conan is not designed to cover.
  - Managing python dependencies of these packages is not something designed. These two packages has dependencies too, they may not break now, but we can not track is in the future
- Pushing the installation step to the upstream via Meson was **not** accepted by both projects and totally understandable. Meson has no native support to install python packages, only to find them.
- Vendorizing the python package in the same recipe, which means, adding the python project to conandata.yml too, then downloading/extracting and configuring to PYTHONPATH and maybe copying to package folder in case needed when running some executable is not a good idea too, as we are not able to track the package version installed as part of the package ID (no workarounds, please).
- The Glib project uses `packaging` basically to compare versions, something very simple that could be done internally. The Conan team will send a propose to avoid all this complexity and something simpler, not requiring `packaging`. This should solve the case by not using packaging anymore there, something useful not only for Conan but also to any user building glib.
- The Mesa project uses Mako only when building the project to generate C files. It can be solved by the simple way of creating a virtual env on-the-fly, then installing mako via pip. It it will be self contained and should not have problems when cross-building, because does not require extra binaries. I'll send PR to https://github.com/conan-io/conan-center-index/pull/20528 with the changes soon. 

I'm closing this PR, as this solution will not be continued.

Issue #559, opened 2024-01-24
Issues building Conan 2 GCC images with Docker Compose


rob-baily:
### Description of Problem, Request, or Question

I have been trying to use the modern path in the repo to build Conan 2 images and have run into a couple problems.

- It seems the base Dockerfile is only set up for Conan 1.0.  The conan install command in there only works in version 1 and not version 2.    It actually seems that it is not needed for Conan 2.0.
- Using the Docker Compose commands results for GCC images in an error in the builder step because there is no gcc-builder target available.

### Environment Details
* Conan Docker Tools image: **N/A**
* Operating System: **Linux**
* Operation System Version: **Ubuntu 20.04.06**  Building for ubuntu18.04
* Python version (If you are building some image): **N/A**

### Steps to reproduce (Include if Applicable)

Run the commands as provided in the documentation for modern: https://github.com/conan-io/conan-docker-tools/tree/master/modern

- _docker-compose build base_ ->This shows the first error for the conan install command.
- _docker-compose build gcc12-builder_ -> This shows the error for the target gcc-builder.

### Build logs (Include if Available)

Will add later. I have resolved these locally so don't have the logs handy but I will go back and get them.

uilianries:
@rob-baily Thank you for reporting, indeed the docker-compose is not used to build official images, we have been using docker commands directly with Jenkins. In case you have some idea about fix it, please, feel free to send a pull request.

rob-baily:
@uilianries Please review my PR and see what you think.  I am not actually clear on how the Conan 2 images are created today as it did not work for me to just update the CONAN_VERSION env.

PR #560, opened 2024-01-24
Build gcc conan2

rob-baily:
Changelog: Fix: #559 Issues building Conan 2 GCC images with Docker Compose

**ADDITONAL NOTES:**

- Changed the compose file for the correct target to use as it seems the Jenkins command uses "builder" as the target
- Added .gitattributes for those that clone to a Windows machine to start
- Update the version in the .env file to Ubuntu 18 as other docs indicate this is the standard.

I verified using these commands from the modern directory for a version of 1.x and 2.x:

- docker-compose build base
- docker-compose build gcc12-builder
- docker-compose build gcc12


modern/.env
```diff
@@ -25,7 +25,7 @@ CLANG12_VERSION=12.0.0
 CLANG13_VERSION=13.0.0
 CLANG14_VERSION=14.0.0
 
-DISTRO=ubuntu16.04
-DISTRO_VERSION=16.04
+DISTRO=ubuntu18.04
```
uilianries, line 28: Please, revert it. Ubuntu 16.04 should be used as main version. The 18.04 is not used by ConanCenterIndex. In the future we should move to 20.04 or 22.04 directly, so we will skip 18.04.

.gitattributes
uilianries, line 1: Yes, it makes sense. Thank you.

modern/base/conan_bootstrap.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -ex
+
+version=`conan --version | awk '{print $3}'`
+
+if [[ "${version}" == 2.* ]]
+then
+    pip install conan==${CONAN_VERSION}
```
uilianries, line 9: The idea was "eating our own dog food", by installing some tools from Conan Center, instead of from APT repository. Here we have binutils installed in /opt. 

So, you can use this bootstrap, but it needs to install for both versions.

uilianries (commented):
Thank you!

modern/base/conan_bootstrap.sh
uilianries: I would suggest reverting it and using only Conan 1.x to install binutils package with lockfile for now. We can update it in the future, when dropping Conan 1.x support.

modern/.env
rob-baily: 2 things on this:

1. I thought according to this https://github.com/conan-io/conan-docker-tools/tree/master/modern#ubuntu-1804-bionic-as-default-distro-october-2022 that  Ubuntu 18.04 was he base distro.  Is that incorrect?
2. Is the .env file used in the Jenkins process that generates images for ConanCenterIndex?  I would think not since the CONAN_VERSION is 2.0.17 and it seems you are still using 1.0 for central.

modern/base/conan_bootstrap.sh
rob-baily: I tested again after reverting these changes and it worked.  I don't know what happened with this step the first time ...

uilianries: By default, it will have Conan 1.x installed, then it installs binutils. As last step, we install CONAN_VERSION. To make it compatible with Conan 2.x would require a new profile file, a new lockfile, a new command line. Too much and not needed for now.

modern/.env
uilianries: > I thought according to this [master/modern#ubuntu-1804-bionic-as-default-distro-october-2022](https://github.com/conan-io/conan-docker-tools/tree/master/modern?rgh-link-date=2024-01-25T13%3A20%3A25Z#ubuntu-1804-bionic-as-default-distro-october-2022) that Ubuntu 18.04 was he base distro. Is that incorrect?

Yes, that file is incorrect. When we moved to 18.04 we actually had a problem with packages already available in Conan Center, built with 16.04. The problem is glibc version, each Ubuntu version has a newer version. So, you can use an older in glibc version in a newer distro, but not the opposite. It resulted in a breakage in Conan Center, so we reverted to 16.04. That document is wrong and need an update.

> Is the .env file used in the Jenkins process that generates images for ConanCenterIndex? I would think not since the CONAN_VERSION is 2.0.17 and it seems you are still using 1.0 for central.

Jenkins uses .env partially, the CONAN_VERSION it consumes directly from build parameters, like, when a new Conan client is released, the CI triggers another job to Jenkins build the docker images, it passes the Conan version released as well. You take a look in the `.ci` folder, it contains all Jenkinsfiles used.

uilianries (approved):
LGTM

uilianries:
@rob-baily Thank you for your fix!

PR #564, opened 2024-02-22
Reduce the CI build

### LINKED ISSUE #504: Proposal to Update GCC Minor Versions

### LINKED ISSUE #563: Build only those images consumed by CCI

Building Docker images is consuming several ours in the CI, but the mainly cost is producing images that are not consumed by ConanCenterIndex.

It would be faster if we could build only what is listed in CCI:

- https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml
- https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v2.yml

In summary:

Legacy images: GCC5, GCC7 and GCC9
Modern images: GCC11 and Clang13

uilianries:
Following the CCI configuration, we are gonna drop the follow configuration:

- GCC 10, 12, 13
- Clang 11, 12, 14

The libstdc++ used in Clang docker images, is copied from GCC 10 docker images. As we will no longer build that version, we will need to copy that library from an image available in hub.docker instead of using from Artfactory.

Plus, minor versions of the used compilers are now updated. It follows #504.

closes #563
closes #504 


modern/clang/Dockerfile
```diff
@@ -247,7 +247,7 @@ RUN apt-get -qq update \
     && cp -a lib/clang/${CLANG_VERSION}/include /tmp/install/lib/clang/${CLANG_VERSION}/include \
     && cp $(find lib -name "*.so*") /tmp/install/lib
 
-FROM ${DOCKER_USERNAME}/gcc${LIBSTDCPP_MAJOR_VERSION}-${DISTRO}${SUFFIX}:${DOCKER_TAG} as libstdcpp
+FROM docker.io/conanio/gcc${LIBSTDCPP_MAJOR_VERSION}-${DISTRO}:latest as libstdcpp
```
danimtb, line 250:
```suggestion
FROM docker.io/${DOCKER_USERNAME}/gcc${LIBSTDCPP_MAJOR_VERSION}-${DISTRO}:latest as libstdcpp
```

uilianries: we can't use DOCKER_USERNAME because it will use Artifactory instead. We do a push to Artifactory for cache images, with names like `docker push center-c3i-docker.jfrog.io/gcc10-ubuntu16.04-builder:master-n23`, then, we push to hub.docker.com with the final.

danimtb: but the value of DOCKER_USERNAME is `conanio`, right? So that will be `dcoker.io/conanio`

uilianries: @danimtb please, take a look in #565

PR #565, opened 2024-02-22
New Jenkins agent version

uilianries:
Dockerhub repository is now a separated parameter, so we can customize it as we need when building locally.

The Jenkins agent is now bumped. Still needs to update the CI interface to follow the same versions.

In the past, we have two different Jenkins agents being built in the same CI build. It was caused because we needed to update the Jenkins agent in the images but there was more involved like the JDK version and the agent script. This is no longer needed, as we can pass by parameter the jenkins client that we need. 

With this new change, any jenkins image will be like: `conanio/gcc<major>-jenkins:<conan_version>-<jenkins_agent_version>`


modern/.env
```diff
@@ -1,5 +1,6 @@
 CONAN_VERSION=2.0.17
 DOCKER_TAG=2.0.17
+DOCKERHUB_USERNAME=conanio
```
danimtb, line 3: Is this `conanio` or `docker.io/conanio`? Sorry, I am still trying to understand 😅

uilianries: Yes, by default it uses docker.io in case passing nothing. It's because when using Artifactory, we actually pass the `center-c3i-docker.jfrog.io`.

uilianries: After merging this PR, I should create a new one to remove this switch between the jenkins agent version. I was a workaround, because in the past we didn't have a way to identify the jenkins agent version, but now we use the image tag.

uilianries:
closing in favor of #566

PR #566, opened 2024-02-23
Add extra parameter for dockerhub repository

uilianries:
This PR is the exactly same as #565 but using the upstream to push commits, so I can really validate changes in the Jenkinsfile. Otherwise, the CI will run what is in the master branch instead.

Dockerhub repository is now a separated parameter, so we can customize it as we need when building locally.

The Jenkins agent is now bumped. Still needs to update the CI interface to follow the same versions.

In the past, we have two different Jenkins agents being built in the same CI build. It was caused because we needed to update the Jenkins agent in the images but there was more involved like the JDK version and the agent script. This is no longer needed, as we can pass by parameter the jenkins client that we need.

With this new change, any jenkins image will be like: conanio/gcc<major>-jenkins:<conan_version>-<jenkins_agent_version>


uilianries:
The current error is related to tests involving Jenkins images. I'll investigate, but my guess it's mixing cached and outdated docker images with the new one.

danimtb (approved):

PR #567, opened 2024-02-26
Do not support x86 in active legacy images

uilianries:
The only still active legacy GCC images: GCC 5, 7 and 9 are installing x86 support, but it's not used in CCI since few years.

This PR removes those steps that 32-bit libraries and tools.

Plus, it splits the Docker recipe in several steps. It will result in a bit bigger image, but will be much easier to debug when failing to build. Errors occurs with frequency, or because a networking error, or an outdated certified, etc.

Plus, fixed the docker-compose file with not supported builds (This file is not used by the CI).

It should be only merged after the PR #566

### DIFF
--- a/legacy/docker-compose.yml
+++ b/legacy/docker-compose.yml
@@ -568,41 +568,6 @@ services:
         image: "${DOCKER_USERNAME}/msvc15:${DOCKER_BUILD_TAG}"
         container_name: msvc15
         tty: true
-    android-clang8:
-        build:
-            context: android-clang_8
-            dockerfile: Dockerfile
-            args:
-                CONAN_VERSION: ${DOCKER_BUILD_TAG}
-        image: "${DOCKER_USERNAME}/android-clang8:${DOCKER_BUILD_TAG}"
-        container_name: android-clang8
-        tty: true
-    android-clang8-x86:
-        build:
-            context: android-clang_8-x86
-            dockerfile: Dockerfile
-            args:
-                CONAN_VERSION: ${DOCKER_BUILD_TAG}
-        image: "${DOCKER_USERNAME}/android-clang8-x86:${DOCKER_BUILD_TAG}"
-        container_name: android-clang8-x86
-        tty: true
-    android-clang8-armv7:
-        build:
-            context: android-clang_8-armv7
-            dockerfile: Dockerfile
-            args:
-                CONAN_VERSION: ${DOCKER_BUILD_TAG}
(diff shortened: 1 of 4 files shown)


legacy/docker-compose.yml
```diff
@@ -568,80 +568,10 @@ services:
         image: "${DOCKER_USERNAME}/msvc15:${DOCKER_BUILD_TAG}"
         container_name: msvc15
         tty: true
-    android-clang8:
```
uilianries, line 571: Removed android because we no longer build it for CCI and not all folders are available, so docker-compose command fails to parse this file.

legacy/docker-compose.yml
```diff
@@ -568,80 +568,10 @@ services:
...
-            dockerfile: Dockerfile
-            args:
-                CONAN_VERSION: ${DOCKER_BUILD_TAG}
-        image: "${DOCKER_USERNAME}/android-clang14-armv8:${DOCKER_BUILD_TAG}"
-        container_name: android-clang14-armv8
     # Jenkins Images
     gcc46-jenkins:
         build:
-            context: jenkins-jenkins
+            context: jenkins
```
uilianries, line 574: typo fix.

legacy/gcc_9/Dockerfile
```diff
@@ -24,11 +23,10 @@ RUN dpkg --add-architecture i386 \
        binutils \
        wget \
        git \
-       libc6-dev-i386 \
        libc6-dev \
-       linux-libc-dev:i386 \
-       g++-9-multilib \
```
uilianries, line 30: multilib installs both x86 and x64 libraries. We no longer build 32-bits in CCI.

legacy/gcc_5/Dockerfile
```diff
@@ -104,6 +104,11 @@ RUN dpkg --add-architecture i386 \
        && apt-get -qq update \
        && rm -rf /var/lib/apt/lists/*
 
+ARG CONAN_VERSION
```
uilianries, line 107: Moved CONAN_VERSION argument, because it represents a layer for Docker. So when we change the Conan version, we actually say to re-build every step after the parameter. However, we just want to install a new Conan version and nothing more, most of time.

danimtb (approved):

danimtb:
Awesome job reviewing this images, may things to remove and update! 👏

RubenRBS (approved):

PR #584, opened 2024-05-02
gcc: pass "--as-needed" when linking

jcar87:
Drop a file called `specs` in the directory where `gcc` looks for it, such that the the file passes `--as-needed` when linking, as it is configured in Ubuntu.

The directory was derived by calling:
```
gcc -print-search-dirs \
  | sed -n -e '/^libraries: =/ {s/.*: =//; s/:/\n/g; p}' \
  | xargs realpath -q \
  | awk '{if (!seen[$0]) print; seen[$0] = 1;}'
```
as instructed in https://wozniak.ca/blog/2024/01/09/1/

The specfile is a copy of `gcc -dumpspecs`, but with the necessary changes to pass `--as-needed` when linking. Even after this change, the output of `gcc -dumpspecs` can be compared/diffed with the contents of this file.

More on `--as-needed`: https://bnikolic.co.uk/blog/gnu-ld-as-needed.html

This fixes issues in Conan Center CI - and avoids the linker encoding `DT_NEEDED` entries for libraries where symbols are not required.

Only doing this for `gcc11` as it is the one currently used with Conan v2 - I wouldn't bother with older versions in Conan 1.x (the problems are unlikely to happen with the Conan 1.x CI due to the behaviour of linking static libraries into shared ones).

### DIFF
--- a/modern/gcc/specs
+++ b/modern/gcc/specs
@@ -1,0 +1,145 @@
+*asm:
+%{m16|m32:--32}  %{m16|m32|mx32:;:--64}  %{mx32:--x32}  %{msse2avx:%{!mavx:-msse2avx}}
+
+*asm_debug:
+%{%:debug-level-gt(0):%{gstabs*:--gstabs;:%{g*:--gdwarf2}}} %{fdebug-prefix-map=*:--debug-prefix-map %*}
+
+*asm_debug_option:
+
+
+*asm_final:
+%{gsplit-dwarf: 
+       objcopy --extract-dwo 	 %{c:%{o*:%*}%{!o*:%w%b%O}}%{!c:%U%O} 	 %b.dwo 
+       objcopy --strip-dwo 	 %{c:%{o*:%*}%{!o*:%w%b%O}}%{!c:%U%O}     }
+
+*asm_options:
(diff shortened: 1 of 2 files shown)


modern/gcc/specs
```diff
@@ -0,0 +1,145 @@
...
+*sysroot_hdrs_suffix_spec:
+
+
+*self_spec:
+
+
+*cc1_cpu:
+%{march=native:%>march=native %:local_cpu_detect(arch %{!m32:64;:32})   %{!mtune=*:%>mtune=native %:local_cpu_detect(tune %{!m32:64;:32})}} %{mtune=native:%>mtune=native %:local_cpu_detect(tune %{!m32:64;:32})}
+
+*link_command:
```
RubenRBS, line 142: From https://gcc.gnu.org/onlinedocs/gcc/Spec-Files.html I get the idea that we could be defining only this, leaving the rest to use the built-in ones?

jcar87: The changes are un `link`, `libgcc` and `link_command`, the latter being the biggest portion of the file. If this file 100% overrides the spec, I would probably leave it as is: it can be diffed with the built-in spec so we know exactly what the changes are, we are only doing it for gcc11, and we have plans in the future to not need this moving forward.

I wouldn't necessarily want to experiment with which specs are defined first by gcc or not if this approach works

uilianries (approved):
Checked with @jcar87, tested locally and it should work for both Conan 1.x and 2.x.

Failed branch CI, but is expected because was pushed directly to CDT repository.

jcar87:
Note: we should only use this file for gcc 11.4 :D

PR #598, opened 2024-06-24
Update Java JDK to version 17

uilianries:
In order to use new Jenkins agents versions, we need to align the JDK version used by the upstream:

https://github.com/jenkinsci/docker-agent/blob/3248.v65ecb_254c298-1/archlinux/Dockerfile#L34

Plus, to avoid collision, we added the JDK version to docker tags too:

`conanio/clang13-ubuntu16.04-jenkins:2.4.1-jdk17-3248.v65ecb_254c298-1`


uilianries:
Current error is because the docker image that hosts GCC 9 (Ubuntu eoan 19.10) is EOL and has no support for JDK 17. When adding `ppa:openjdk-r/ppa`, the JDK 14 is latest version available.

I'll try to install using another distro and checking the compatibility.

uilianries:
Used Xenial (16.04) as base for installing OpenJDK in Eoan (19.10). So, it worked locally.

uilianries:
The Jenkins agent version does not use the suffix `-1` or whatever. It marks a change in the repository, but the agent version is always the same: https://github.com/jenkinsci/docker-agent/blob/3248.v65ecb_254c298-6/archlinux/Dockerfile#L70

PR #601, opened 2024-07-04
ci: Build with Ubuntu 22.04

uilianries:
Generate modern images using Ubuntu 22.04 to be consumed by Github Action when building clion plugin.


uilianries:
Closing after finishing testing. It's possible to update to Ubuntu 22.04 without much effort.

PR #604, opened 2024-07-13
Return CONAN_VERSION arg to modern .env

CommanderCRM:
Changelog: (Fix): This env variable deletion made `docker compose build base` impossible in modern folder, as it is used during the build process

- [X] I've read the [Contributing guide](https://github.com/conan-io/conan-docker-tools/blob/master/.github/CONTRIBUTING.md).
- [X] I've followed the [PEP8](https://www.python.org/dev/peps/pep-0008) style guides for Python code.
- [X] I've followed the [Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) guides for Dockerfile.


uilianries (approved):
LGTM. Indeed we don't use docker compose to generate images. It has a limitation related to cache-from parameters.

PR #633, opened 2025-06-09
Install Python 3.12 as optional version

uilianries:
Having Python 3.12 is important to run internal tools in the CI, but as Conan should be working with an older version, both Python versions can be installed in parallel in the image.

The OpenSSL 1.1 is required in order to build Python 3.12: https://github.com/pyenv/pyenv/wiki/Common-build-problems#2-your-openssl-version-is-incompatible-with-the-python-version-youre-trying-to-install; Ubuntu 16.04 only offers OpenSSL 1.0, so we need to build it from source. 

UPDATE: OpenSSL 3.x is compatible and will be used instead of 1.1.x version.


images/base/Dockerfile
```diff
@@ -92,6 +93,23 @@ RUN wget --no-check-certificate --quiet -O /tmp/pyenv-installer https://raw.gith
     && ln -s /usr/share/pyshared/lsb_release.py /opt/pyenv/versions/${PYTHON_VERSION}/lib/python${PYTHON_VERSION%.*}/site-packages/lsb_release.py \
     && pip install -q --upgrade --no-cache-dir pip
 
+# Python +3.10 requires OpenSSL 1.1.x, but Ubuntu 16.04 LTS ships with OpenSSL 1.0.x
+RUN wget --no-check-certificate --quiet -O /tmp/openssl-1.1.1w.tar.gz https://github.com/openssl/openssl/releases/download/OpenSSL_1_1_1w/openssl-1.1.1w.tar.gz \
```
jcar87, line 97: does it work with a recent OpenSSL 3.x? as 1.x is already EOL

uilianries: Good question! Let me try with OpenSSL 3.x. Thank you for reviewing!

jcar87 (approved):
Thanks!!

